PR #2, opened 2021-11-30
Up to date spectre-meltdown-checker fixes

nobodysu:
- Going further, since `find` is very noisy, I think some permissions on `/proc` need to be relaxed - nothing serious, just a directory traverse. Otherwise it can shadow other DENIALs like with me earlier. Why not `deny` directive? If some other functionality will be added to `spectre-meltdown-checker` debugging on our side will be rather painful.
- Without the modules section two mitigations were false negative.
- `ptrace` is used by readlink and does not affect mitigations detection, at least at first glance.
```
apparmor="DENIED" operation="capable" profile="spectre-meltdown-checker" pid=1356 comm="readlink" capability=19  capname="sys_ptrace"
apparmor="DENIED" operation="ptrace" profile="spectre-meltdown-checker" pid=1357 comm="readlink" requested_mask="read" denied_mask="read" peer="unconfined"
```
Remove of you don't like it.

### DIFF
--- a/apparmor.d/profiles-s-z/spectre-meltdown-checker
+++ b/apparmor.d/profiles-s-z/spectre-meltdown-checker
@@ -6,7 +6,7 @@ abi <abi/3.0>,
 
 include <tunables/global>
 
-@{exec_path} = /{usr/,}bin/spectre-meltdown-checker
+@{exec_path} = /{,usr/}{,local/}bin/spectre-meltdown-checker{,.sh}
 profile spectre-meltdown-checker @{exec_path} {
   include <abstractions/base>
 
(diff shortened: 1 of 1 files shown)


apparmor.d/profiles-s-z/spectre-meltdown-checker
```diff
@@ -154,6 +163,11 @@ profile spectre-meltdown-checker @{exec_path} {
   profile kmod {
     include <abstractions/base>
 
+    capability sys_module,
+
+    owner /sys/module/cpuid/** r,
```
roddhjav, line 168: Can you use `@{sys}` instead of `/sys`

nobodysu: Sure.

roddhjav:
Thank for the PR. You are right, the script require extended access over `/proc/*/`
Merged.

PR #3, opened 2021-11-30
QEMU guest virtual disks compatibility

nobodysu:
Not all paths need to be changed since QEMU is using the drives through it's own driver, so hardware-specific commands will not work anyway.


roddhjav:
Thank. Merged.

PR #9, opened 2021-12-12
bind-utils

nobodysu:


apparmor.d/profiles-a-f/dig
```diff
@@ -24,5 +24,9 @@ profile dig @{exec_path} {
 
   owner @{HOME}/.digrc r,
 
+  /tmp/batch_mode.dig       r,
+  /home/dig/batch_mode.dig  r,
```
roddhjav, line 28: What is this `/home/dig` directory?

roddhjav:
Thank for your PR.

On top of my comment in the code, can you change the rules order in `host` and `nslookup`?
You should take inspiration of the existing profile where you usually have this order:
- include
- network 
- @{exec_path} r,
- the other file based rules...

I am sorry to complain about this, but as they are a lot of profile, it is easier to maintain them whem they all share a common structure.

nobodysu:
Of course. I didn't quite caught the style yet.

apparmor.d/profiles-a-f/dig
nobodysu: A supposed place for files which should only be accessible by dig.

nobodysu:
> @{exec_path} r,

Could you reconsider? This directive is needed almost every time and will benefit from moving it out of sight. E.g. to the similar directives right before the includes, for locality.

nobodysu:
> The binaries and library required: `/{usr/,}bin/`, `/{usr/,}lib/`, `/opt/`...

Empty option (after the last comma) should go first, in my opinion, otherwise it's more prone to error.
When empty option is mentioned as a prefix - it stays put, and less likely to be removed accidentally when changing the list of values in the future.

Also I wonder, how dangerous is the `m` directive on a binary? It seems some distributives mandates it (e.g. Armbian), while others don't.

roddhjav:
Most of the "styles" actually come from [apparmemall](https://gitlab.com/morfikov/apparmemall) and its about 600 profiles. So when I added my own 600 profiles, I kept the same style in order to integrate them correctly with what was already done. `@{exec_path} r,` and `/{usr/,}` come from there. So no, I won't change it, because that simply too much work.

Also, it is common to start the profile by the inclusions. 

> Also I wonder, how dangerous is the m directive on a binary?

A lot of program need it to work. So that will change depending of the profile.

apparmor.d/profiles-a-f/dig
roddhjav: I don't get it. `/home/dig` implies there is a `dig` user on the system. Is this your own config. If so, it should go in a `local/dig` file.

nobodysu: What about now?

roddhjav:
Thank for the update, merged.

PR #11, opened 2021-12-12
diff and colordiff

nobodysu:


roddhjav:
Thank for this PR. However, I will not merge it because these tiny tools should not be confined by themselves. It would go the same with profiles for software like `ls`, `rm`, `cat`...  Let me explain.

These are general tools that in a general context can legitimately access any file in the system. And such the confinement of such tools by a global profile would be minimal and may I say, useless.
For example, with this rule, you basically allow to diff any file in the system:
```
@{DIFF_PATHS_RO} = /etc @{HOME} /srv /tmp /var /usr/share /usr/lib/systemd
...
@{DIFF_PATHS_RO}/** r,
```

It get even worse as for example, `/etc` can contains sensitive file, you now you explicitly deny some `extremely sensitive files`. But how do you know the exhaustive list of *sensitive files* in `/etc`, `/tmp`, or `/var? This breaks the principle of mandatory access control that is to only allow what is required. Here you allow everything and blacklist some paths.

It creates even more issue when you want to use these profile in other profile: `/{,usr/}bin/diff rPx,` will restrict access to the very generic and not very confined diff profile. Whereas most of the time, you want restrict diff to some file specific to your profile.

**Some example of integration of `diff` in other profiles:**
* In `dpkg`, a child profile (`rCx -> diff`), allows `diff` to only access etc config files:
https://github.com/roddhjav/apparmor.d/blob/accf5538bdfc1598f1cc1588a7118252884df50c/apparmor.d/groups/apt/dpkg#L123
* In `pass`, as it is a dependency of pass. Here `diff` inherit pass profile and has the same access than the pass profile, so it will be allowed to diff password files because more than a generic `diff` it is a `diff` for the pass password manager:
https://github.com/roddhjav/apparmor.d/blob/accf5538bdfc1598f1cc1588a7118252884df50c/apparmor.d/profiles-m-r/pass#L20

nobodysu:
Alright, I understand.
> But how do you know the exhaustive list of sensitive files in /etc, /tmp, or /var?

I followed the convention when all keys are common to be in `/etc`.

> And such the confinement of such tools by a global profile would be minimal and may I say, useless.

Even when process can read all system files, confinement can still be very beneficial in case of malicious file substitution, code injection or reporting to the C&C server.
Now when you mention it, I understand that `diff` was approached poorly. As for `colordiff` and `grc`, I firmly believe they need to be done in some form or another. While `colordiff` have relatively medium risk, python-based `grc` can have tremendous impact in case some [popular library will be infiltrated](https://www.zdnet.com/article/twelve-malicious-python-libraries-found-and-removed-from-pypi/).
All these programs are dealing with arbitrary user input, and in my opinion should be confined. Due to their nature it could be a blacklist way.

roddhjav:
> All these programs are dealing with arbitrary user input, and in my opinion should be confined.

The solution here is sandboxing, not confinement. For cli tool, you should use something like [toolbox](https://github.com/containers/toolbox). Have a look at the [project concepts](https://github.com/roddhjav/apparmor.d#concepts) to see the logic behind this. 

> Due to their nature it could be a blacklist way.

Blacklisting breaks the principle of mandatory access control. But is fine in app sandboxing.

PR #18, opened 2021-12-24
Update acpid

nobodysu:
```
debian11$ grep -E '^\s+/dev/null' /etc/apparmor.d/abstractions/base
  /dev/null                      rw,
debian11$ find /usr/*bin -name acpid
/usr/sbin/acpid
```

```
ubuntu-lts$ grep -E '^\s+/dev/null' /etc/apparmor.d/abstractions/base
  /dev/null                      rw,
ubuntu-lts$ find /usr/*bin -name acpid
/usr/bin/acpid
```

### DIFF
--- a/apparmor.d/profiles-a-f/acpid
+++ b/apparmor.d/profiles-a-f/acpid
@@ -6,7 +6,7 @@ abi <abi/3.0>,
 
 include <tunables/global>
 
-@{exec_path} = /{usr/,}bin/acpid
+@{exec_path} = /{usr/,}{,s}bin/acpid
 profile acpid @{exec_path} flags=(attach_disconnected) {
   include <abstractions/base>
   include <abstractions/nameservice-strict>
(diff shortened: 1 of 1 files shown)


roddhjav:
Thank a lot for this PR. And good work too. This profile was missing a bit of love.

Can you just apply the profile guideline to it.

apparmor.d/profiles-a-f/acpid
```diff
@@ -20,17 +20,68 @@ profile acpid @{exec_path} flags=(attach_disconnected) {
...
+
+  /etc/acpi/powerbtn-acpi-support.sh r,
+
+  /{usr/,}bin/sed          rix,
+  /{usr/,}bin/pgrep        rix,
+  /{usr/,}bin/{,e}grep     rix,
+  /{usr/,}bin/pinky        rix,
+  /{usr/,}bin/{,ba,da}sh   rix,
+  /{usr/,}bin/dbus-send    rix,
+  /{usr/,}bin/systemctl    rix,
```
roddhjav, line 57: Here you should use `/{usr/,}bin/systemctl rPx -> child-systemctl, instead

apparmor.d/profiles-a-f/acpid
```diff
@@ -20,17 +20,68 @@ profile acpid @{exec_path} flags=(attach_disconnected) {
...
+  /{usr/,}sbin/killall5    rix,
```
roddhjav, line 58: Some distributions do not have sbin directory (eg: arch). So you should use `{s,}bin`.

roddhjav:
Thank for the update, it is better now, once you fix the minor issue I can merge it.

nobodysu:
Done.

apparmor.d/profiles-a-f/acpid
```diff
@@ -17,20 +17,63 @@ profile acpid @{exec_path} flags=(attach_disconnected) {
 
   @{exec_path} mr,
 
-  /{usr/,}bin/{,ba,da}sh        rix,
+  /{usr/,}bin/{ba,da,}sh        rix,
   /{usr/,}bin/logger            rix,
 
+  /etc/acpi/powerbtn-acpi-support.sh rPx -> acpid//powerbtn-acpi-support.sh,
```
nobodysu, line 23: Double slash or underscore here? Slash will collide with `rCx` visually.

PR #20, opened 2022-01-16
Update pulseaudio

nobodysu:


apparmor.d/profiles-m-r/pulseaudio
```diff
@@ -66,18 +69,72 @@ profile pulseaudio @{exec_path} {
   @{sys}/devices/virtual/dmi/id/{bios_vendor,board_vendor,sys_vendor} r,
   @{sys}/devices/system/node/ r,
   @{sys}/devices/system/node/node[0-9]/meminfo r,
-
-  owner @{run}/user/@{uid}/dconf/ w,
-  owner @{run}/user/@{uid}/dconf/user rw,
+  @{sys}/module/apparmor/parameters/enabled r,
```
roddhjav, line 72: Is this really needed?

apparmor.d/profiles-m-r/pulseaudio
```diff
@@ -66,18 +69,72 @@ profile pulseaudio @{exec_path} {
...
+
+  owner @{run}/user/@{uid}/dconf/       rw,
+  owner @{run}/user/@{uid}/dconf/user   rw,
+  owner @{run}/user/@{uid}/ICEauthority r,
+  owner @{run}/user/@{uid}/systemd/notify rw,
 
   owner @{PROC}/@{pid}/fd/ r,
   owner @{PROC}/@{pid}/stat r,
 
-  /var/lib/dbus/machine-id r,
```
roddhjav, line 79: Did you remove machine-id?

nobodysu: It's in abstraction now:
```
$ grep machine-id abstractions/dbus*
abstractions/dbus-session-strict:  /etc/machine-id r,
abstractions/dbus-session-strict:  /var/lib/dbus/machine-id r,
```

apparmor.d/profiles-m-r/pulseaudio
nobodysu: It seems like pulseaudio is AppArmor-aware. We could block it, but I'm not sure about side-effects,I haven't dug the code. The sound is present.

roddhjav:
Thank for the PR. Merged.

PR #22, opened 2022-01-19
systemd-timesyncd.service

nobodysu:
Needed for at least Ubuntu.


nobodysu:
Question: why not specify all these rules in single file - `apparmor.service`?

roddhjav:
I simply followed what other projects do to enforce order for this late aa profile load.

Thank, merged.

PR #37, opened 2022-05-22
sshd: Ubuntu compatibility

nobodysu:


apparmor.d/groups/ssh/sshd
```diff
@@ -73,17 +76,17 @@ profile sshd @{exec_path} flags=(attach_disconnected) {
         @{run}/motd.dynamic.new rw,
         @{run}/resolvconf/resolv.conf r,
         @{run}/systemd/sessions/[0-9]*.ref rw,
-        @{run}/systemd/userdb/ r,
```
nobodysu, line 76: Already present in `nameservice-strict`

apparmor.d/groups/ssh/sshd
```diff
@@ -73,17 +76,17 @@ profile sshd @{exec_path} flags=(attach_disconnected) {
...
   @{sys}/fs/cgroup/*/user/*/[0-9]*/ rw,
   @{sys}/fs/cgroup/systemd/user.slice/user-@{uid}.slice/session-c[0-9]*.scope/ rw,
 
-  owner @{PROC}/@{pid}/limits r,
-  owner @{PROC}/@{pid}/loginuid rw,
-  owner @{PROC}/@{pid}/mounts r,
-  owner @{PROC}/@{pid}/oom_adj rw,
-  owner @{PROC}/@{pid}/oom_score_adj rw,
-  owner @{PROC}/@{pid}/uid_map r,
+  owner @{PROC}/@{pids}/limits r,
```
nobodysu, line 84: We are switching to `@{pids}`, right?

roddhjav: No we are not, `@{pid}` and `@{pids}` will in the future not be the same thing. `@{pid}` refers to the pid of the profile. In practice, I usually putt `@{pid}` with `owner` and  `@{pids}` without it.

apparmor.d/groups/ssh/sshd
```diff
@@ -43,14 +43,17 @@ profile sshd @{exec_path} flags=(attach_disconnected) {
...
   ptrace (read,trace) peer=unconfined,
 
+  network inet  stream,
+  network inet6 stream,
+
   @{exec_path} mrix,
 
-  /{usr/,}bin/{,b,d,rb}ash         rUx,
-  /{usr/,}bin/{c,k,tc,z}sh         rUx,
+  /{usr/,}bin/{,b,d,rb}ash         rPUx,
```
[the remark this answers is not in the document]
nobodysu: Alright, I got you.
I haven't experienced any breakage on Debian-derivatives, since `bash` there is always personal and all other process-originating shells uses `dash`.

apparmor.d/groups/ssh/sshd
nobodysu: Got it, reverted.

PR #49, opened 2022-06-02
Ubuntu 22.04, second batch

nobodysu:


roddhjav:
As I have now added support for dbus for most system services, could you test again your modification from the last changes?

nobodysu:
Should I drop `peer=(name=...)`?

roddhjav:
No, I use it myself sometime.

nobodysu:
Could we use one of the following formatting? Because with current one important parameters got visually lost in the indentation.
```
  dbus (receive) bus=session
       path=/org/freedesktop/Notifications interface=org.freedesktop.Notifications member=NotificationClosed peer=(name=:*),
```
or
```
  dbus (receive) bus=session
       path=/org/freedesktop/Notifications
       interface=org.freedesktop.Notifications
       member=NotificationClosed
       peer=(name=:*),
```

apparmor.d/groups/freedesktop/polkitd
```diff
@@ -22,14 +22,30 @@ profile polkitd @{exec_path} {
 
   ptrace (read),
 
-  dbus (send,receive) bus=system path=/org/freedesktop/PolicyKit[0-9]/*
-       interface=org.freedesktop.{DBus.Introspectable,DBus.Properties,PolicyKit[0-9].*},
-
-  dbus send bus=system path=/org/freedesktop/DBus
+  dbus (send) bus=system path=/org/freedesktop/DBus
```
nobodysu, line 25: The idea here is to granularize and allow sending all members, but receive only mentioned.

nobodysu:
What's the purpose of conditional `@{libexec}`? It brings additional complexity and requires a configure script while not covering all cases, e.g.:
```
*symlinks ommited*
xubuntu2204$ ls -l ...
/usr/libexec/polkit-agent-helper-1
/usr/libexec/gvfsd-admin
/usr/lib/update-notifier/package-system-locked
/lib/update-notifier/package-system-locked
```
Isn't `+=` more appropriate for `@{libexec}` tunable?

roddhjav:
I did it to ensure not allowing run from `/usr/lib` when the lib in is `/usr/libexec`. Ultimately it would be possible to have something like `@{libexec}=/{usr/,}lib{,exec}/` but that would require to change all `/{usr/,}lib` by `@{libexec}` which raises even more changes and complexity.

> it requires a configure script

A configure/build script is still needed anyway, so that does not change that much

All lib in your example are actually correctly managed by the current profile.

apparmor.d/groups/freedesktop/polkitd
roddhjav: That's look like a pain to manage all of this for all profiles.

I mean, in concept `polkitd` 'own' the `/org/freedesktop/PolicyKit[0-9]` namespace in dBus. Therefore it makes sense to allow everything in it regardless send or receive. (cf the systemd polkit service definition that does the same).

It is also a simplification as a change in the program would not change the profile.

nobodysu: Alright, that's reasonable. I hope send/receive parenthesis is OK.

roddhjav: As you might have seen, I usually do not put them unless it is mandatory, but honestly we have other priorities than tracking parenthesis.

roddhjav:
Thank, for the PR. Merged!

PR #50, opened 2022-06-04
Update spectre-meltdown-checker

nobodysu:


apparmor.d/profiles-s-z/spectre-meltdown-checker
```diff
@@ -56,11 +57,13 @@ profile spectre-meltdown-checker @{exec_path} {
   /{usr/,}bin/{,@{multiarch}-}strings rix,
   /{usr/,}bin/{,@{multiarch}-}objdump rix,
   /{usr/,}{s,}bin/iucode_tool         rix,
+  /{usr/,}{s,}bin/rdmsr               rix,
```
nobodysu, line 60: https://github.com/speed47/spectre-meltdown-checker/blob/b68ebe67f29dda87c296e806ff3360e13a8f13c6/spectre-meltdown-checker.sh#L2680

nobodysu:
https://github.com/roddhjav/apparmor.d/blob/4d9a5d6c4de825afd4d32d981e0a7c59421a381e/apparmor.d/profiles-s-z/spectre-meltdown-checker#L83
Self-upgrating? I think this should be revisited.

roddhjav:
> Self-upgrating? I think this should be revisited.

You are right, now would it work without it?

nobodysu:
It *is* updating mcedb with current profile. *Self*-updating is needed only when in-place update is called:
```
# The builtin version follows, but the user can download an up-to-date copy (to be stored in his $HOME) by using --update-fwdb
# To update the builtin version itself (by *modifying* this very file), use --update-builtin-fwdb
```
User could just download original script by hand or other means.

PR #57, opened 2022-07-16
Various containerd fixes

Jeroen0494:
Various fixes for containerd. When running Calico, it requires access to a lot of stuff under devices but it seems to only want an index.

### DIFF
--- a/apparmor.d/groups/virt/containerd
+++ b/apparmor.d/groups/virt/containerd
@@ -16,6 +16,7 @@ profile containerd @{exec_path} flags=(attach_disconnected) {
 
   capability chown,
   capability dac_read_search,
+  capability dac_override,
   capability net_admin,
   capability sys_admin,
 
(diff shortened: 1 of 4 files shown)


Jeroen0494:
I do still get the following error:
```
  Warning  FailedCreatePodSandBox  10s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create network namespace for sandbox "71ea90afb508946a1476bfd75c50944cb6412259219080e83ac589fecac2b5d6": failed to setup netns: failed to create namespace: open /proc/2390/task/22986/ns/net: permission denied
```

When I execute `aa-teardown`, everything works as expected. There are no logs for this in the audit log, also I explicitly allow it:
```
@{PROC}/@{pid}/task/@{tid}/ns/net rw,
```

Any idea what could be going wrong?

apparmor.d/groups/virt/containerd
```diff
@@ -72,28 +76,29 @@ profile containerd @{exec_path} flags=(attach_disconnected) {
...
+  /dev/hugepages/      r,
+  /dev/input/          r,
+  /dev/input/by-id/    r,
+  /dev/input/by-path/  r,
+  /dev/net/            r,
+  /dev/snd/            r,
+  /dev/snd/by-path/    r,
+  /dev/vfio/           r,
+
+  deny / r,
```
roddhjav, line 101: I would recommend not denying `/` on a profile that has the `attach_disconnected` flag as `/` can be the root of something else than the host.

Jeroen0494: Good point, I removed it.

PR #58, opened 2022-07-18
Support for k3s

Jeroen0494:
Hi,

This PR adds preliminary support for k3s. This hasn't been tested in production yet, but I do use it in my testing virtual machine. Chances are this will need some more fixes when I've figured out what AppArmor profile causes this error:
```
Warning  FailedCreatePodSandBox  5s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "4c29d2d34e0960610a840fc07d2d3b03664530b31b2192672e4966ce7f9173ae": Failed to open current namespace: open /proc/40621/task/40621/ns/net: permission denied
```

But I do have containers running:
```
# kubectl get pod -A
NAMESPACE          NAME                                       READY   STATUS              RESTARTS   AGE
kube-system        metrics-server-7cd5fcb6b7-tqf8w            0/1     Unknown             0          2d
calico-apiserver   calico-apiserver-6bdc5c6b5c-ksrh9          0/1     Unknown             0          2d
kube-system        local-path-provisioner-6c79684f77-45zhk    0/1     Unknown             0          2d
calico-apiserver   calico-apiserver-6bdc5c6b5c-676j5          0/1     Unknown             0          2d
calico-system      calico-kube-controllers-54fcfcd697-dcbbs   0/1     ContainerCreating   0          47h
tigera-operator    tigera-operator-5f9bdc5c59-ldbj4           1/1     Running             0          135m
calico-system      calico-typha-6886b56548-b594n              1/1     Running             11         2d
calico-system      calico-node-7tffj                          1/1     Running             0          135m
kube-system        coredns-d76bd69b-bwdtb                     0/1     ContainerCreating   0          5m16s
```

This PR also fixes some pkttyagent permission errors, and adds support for the xtables-nft tool which allows using iptables configuration while using the nftables kernel API.

I'm unsure about all of the symbolic links to this command in the k3s profile, basically on Ubuntu 22.04 all iptables commands are redirected to the xtables-nfs tool, so all commands should use this profile as well. But the clutter is obvious and could use a more elegant solution.


apparmor.d/groups/virt/k3s
```diff
@@ -0,0 +1,176 @@
...
+  /var/log/kubelet/{,**}       r,
+  /var/log/kubernetes/{,**}    r,
+  /var/log/kubernetes/audit/** rw,
+  /var/log/pods/{,**}          r,
+  /var/log/pods/{,**/}         rw,
+  /var/log/pods/**/[0-9]*.log  rw,
+
+  @{HOME}/.kube/cache/discovery/{,**}         rw,
+  @{HOME}/.kube/cache/http/[0-9a-z]*          rw,
+  @{HOME}/.kube/cache/http/.diskv-temp/[0-9]* rw,
```
roddhjav, line 100:
```suggestion
  owner @{HOME}/.kube/cache/discovery/{,**}         rw,
  owner @{HOME}/.kube/cache/http/[0-9a-z]*          rw,
  owner @{HOME}/.kube/cache/http/.diskv-temp/[0-9]* rw,
```
Usually you always have `owner @{HOME}` never `@{HOME}` (with some exception of course)

apparmor.d/groups/virt/k3s
```diff
@@ -0,0 +1,176 @@
...
+  capability sys_admin,
+  capability sys_resource,
+
+  ptrace peer=@{profile_name},
+  ptrace (read) peer=unconfined,
+
+  network inet    dgram,
+  network inet6   dgram,
+  network inet    stream,
+  network inet6   stream,
```
roddhjav, line 30:
```suggestion
  network inet dgram,
  network inet6 dgram,
  network inet stream,
  network inet6 stream,
```

Only the is executable rules (rPx, rix...) are (sometime) aligned to the right. In case of doubt, just do not do it.

apparmor.d/groups/network/xtables-nft-multi
```diff
@@ -0,0 +1,36 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2022 Jeroen Rijken
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = /{usr/,}sbin/xtables-nft-multi
```
roddhjav, line 9: Always use `{s,}bin` instead of `sbin` as some distributions do not have `sbin`
```suggestion
@{exec_path} = /{usr/,}{s,}bin/xtables-nft-multi
```

apparmor.d/groups/virt/k3s
```diff
@@ -0,0 +1,176 @@
...
+
+  /{usr/,}{s,}bin/iptables-restore rPx -> xtables-nft-multi,
+  /etc/alternatives/iptables-restore rPx -> xtables-nft-multi,
+  /{usr/,}{s,}bin/iptables-legacy-restore rPx -> xtables-nft-multi,
+  /{usr/,}{s,}bin/xtables-nft-multi rPx,
+
+  /usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds/uds rix,
+  /var/lib/rancher/k3s/data/[0-9a-f]*/bin/* rix,
+
+  /usr/libexec/kubernetes/kubelet-plugins/volume/exec/{,**} r,
```
roddhjav, line 62: Always use `@{libexec}` instead of `/usr/libexec` as some distributions do not have `/usr/libexec`
```suggestion
  @{libexec}/kubernetes/kubelet-plugins/volume/exec/{,**} r,
```

apparmor.d/profiles-m-r/pkttyagent
```diff
@@ -12,15 +13,35 @@ profile pkttyagent @{exec_path} {
...
+       member=BeginAuthentication,
+  
+  dbus receive bus=system path=/org/freedesktop/PolicyKit[0-9]/Authority
+       interface=org.freedesktop.PolicyKit[0-9].Authority
+       member=Changed,
 
   @{exec_path} mr,
 
+  /etc/nsswitch.conf r,
+  /etc/passwd r,
```
roddhjav, line 40: Remove this, add the `nameservice-strict` instead.

roddhjav:
Hi, thank for your PR. As usual, have a look at the comments.

Also usually, `xtables-nft-multi` is handled internally (and differently) by all profile that need it. Therefore there is no one common and generic `xtables-nft-multi` profile that anyone can call when needed. See the [multiple use of xtables-nft-multi](https://github.com/roddhjav/apparmor.d/search?q=xtables-nft-multi) in the project to understand what I mean.

Jeroen0494:
Perhaps my SELinux mindset is playing tricks with me here. Usually with SELinux I would create an interface that other profiles can load to gain certain privileges. AppArmor doesn't really have such a mechanism without switching profiles. So how do I give only temporary elevated privileges without extending them to the entire profile?

k3s is a single binary for all things Kubernetes, and Kubernetes does have components that handle iptables/nftables/eBPF stuff. But with vanilla k8s that's only a single component gaining those privileges. With k3s, it's everything.

Jeroen0494:
How would you feel about a subprofile that gets used when the xtables-nfs-xxxx command is executed? That way there is no generic xtables profile, but privileges are still minimal for the k3s binary itself. Am I right?

roddhjav:
You have a few options:
* A subprofile (but beware, a subprofile must have less privilege than the parent profile), but I think, in this context it is fine.
* Integrate the rule in the main profile
* Create a dedicated profile for this, called `k3s-nft` for example of `child-nft` is the rule can be used for profile.

Jeroen0494:
I managed to solve the `unable to create network sandbox problem` error! Turns out it was a calico permission problem. I'm not sure why this wasn't reported to audit, but I'm glad it's fixed now.

Jeroen0494:
Ready for final review and merge. Not sure why the build fails, the error has already been fixed: https://github.com/Jeroen0494/apparmor.d/blob/feat/k3s/apparmor.d/groups/virt/containerd

apparmor.d/groups/virt/cni-calico
```diff
@@ -26,6 +34,11 @@ profile cni-calico @{exec_path} {
   
   @{run}/calico/ rw,
   @{run}/calico/ipam.lock rwk,
+  @{run}/netns/cni-@{uuid} r,
+
+  @{PROC}/sys/net/ipv4/ip_forward rw,
+  @{PROC}/sys/net/ipv4/{conf,neigh}/cali[0-9a-z]*/* rw,
+  @{PROC}/sys/net/ipv6/{conf,neigh}/cali[0-9a-z]*/* rw,
```
roddhjav, line 41: Usually when there is ipv4, there is ipv6 too:
```suggestion
  @{PROC}/sys/net/ipv{4,6}/ip_forward rw,
  @{PROC}/sys/net/ipv{4,6}/{conf,neigh}/cali[0-9a-z]*/* rw,
```

apparmor.d/groups/virt/containerd
```diff
@@ -36,38 +37,44 @@ profile containerd @{exec_path} flags=(attach_disconnected) {
...
   /opt/cni/bin/bandwidth rPx,
-  /opt/cni/bin/calico    rPx,
+  /opt/cni/bin/calico rPx,
 
   /opt/containerd/{,**} rw,
 
+  /var/lib/cni/{,**/} w,
   /var/lib/cni/results/cni-loopback-@{uuid}-lo wl,
+  /var/lib/cni/results/cni-loopback-[0-9a-z]*-lo wl,
+  /var/lib/cni/results/k8s-pod-network-[0-9a-z]*-eth0 wl,
```
roddhjav, line 66: As the interfaces can have a large set of name, you cannot just use `eth0` here.

apparmor.d/groups/virt/k3s
```diff
@@ -0,0 +1,189 @@
...
+  /var/log/kubelet/{,**} r,
+  /var/log/kubernetes/{,**} r,
+  /var/log/kubernetes/audit/** rw,
+  /var/log/pods/{,**} r,
+  /var/log/pods/{,**/} rw,
+  /var/log/pods/**/[0-9]*.log rw,
+
+  owner @{HOME}/.kube/cache/discovery/{,**} rw,
+  owner @{HOME}/.kube/cache/http/[0-9a-z]* rw,
+  owner @{HOME}/.kube/cache/http/.diskv-temp/[0-9]* rw,
```
roddhjav, line 85: Same here, why not `owner @{HOME}/.kube/cache/{,**} rw,`?

apparmor.d/groups/virt/containerd
Jeroen0494: These interface names are created and managed by Calico. I haven't found any documentation yet on whether these names are adjustable, or different on systems where eth* is used by the host. Calico does have the capability to manage host interfaces as well. it just so happens to be used by Kubernetes so much as an CNI. But I don't really know what other interface names I can expect.

Perhaps something like this could work:
`/var/lib/cni/results/k8s-pod-network-[0-9a-z]*-* wl,`

apparmor.d/groups/virt/k3s
```diff
@@ -0,0 +1,189 @@
...
+  /var/lib/kubelet/{cpu_manager_state,memory_manager_state} r,
+  /var/lib/kubelet/device-plugins/{,DEPRECATION,kubelet.sock} rw,
+  /var/lib/kubelet/pod-resources/{kubelet.sock,[0-9]*} rw,
+  /var/lib/kubelet/pods/@{uuid}/containers/*/[0-9a-f]* rw,
+  /var/lib/kubelet/pods/@{uuid}/etc-hosts rw,
+  /var/lib/kubelet/pods/@{uuid}/plugins/kubernetes.io~*/{,**} rw,
+  /var/lib/kubelet/pods/@{uuid}/volumes/kubernetes.io~*/{,**} rw,
+  /var/lib/kubelet/pods/@{uuid}/**/ca.crt rw,
+  /var/lib/kubelet/pods/@{uuid}/**/namespace rw,
+  /var/lib/kubelet/pods/@{uuid}/**/token rw,
```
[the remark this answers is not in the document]
Jeroen0494: The idea is to only give permissions to the strictly necessary. These rules are enough for my deployments for now, write access to the whole kubelet folder is not really needed at this point. If it does prove necessary, I can always add rules, or replace them with the solution you suggest.

apparmor.d/groups/virt/k3s
Jeroen0494: This could be simplified indeed.

Jeroen0494:
Thanks for the comments, I've made some adjustments.

I plan on using this profile (and the containerd-shim) myself on my server. Right now there profiles are made from a simple k3s cluster in a VM. My own server runs a Nextcloud stack, some download tools and Mopidy for music. That should give it a proper review. I just want to make sure the fixes for calico and containerd are in before I run this on something I depend on. I can (and will) always make adjustments to these profiles later on.

apparmor.d/groups/virt/k3s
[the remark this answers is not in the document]
Jeroen0494: So after performing a clean installation of k3s, I noticed it is not the installation script that performs the setup of the `/var/lib/kubelet` directory, but the k3s process itself. Now it is more logical to apply your suggestion. I'll update the PR.

Jeroen0494:
I'm going to need xtables-nfs-xxxx not just for k3s but also for cni-portmap. I think it'd be best to create a separate cni-xtables-nfs profile and load this from k3s and cni-portmap.

Jeroen0494:
Done, the profile is now separate. The path to the executable is not in the profile name, so it shouldn't conflict with other profiles.

Jeroen0494:
Ready for another review (sans the Calico interface name, which I'm looking in to).

roddhjav:
Thank a lot for this PR. Merged.

PR #64, opened 2022-08-13
Support for Grub and Avahi, general update

Jeroen0494:
Hi,

* Add support for all Grub commands
* Add support for all Avahi commands
* Add support for whoami, whereis, which, findmnt, users, sanoid and syncoid
* Replace various /dev/tty statements with consoles abstraction
* Various CNI and ZFS fixes
* General update for a lot of profiles

The grub-mkconfig path has the same exact issue with privilege escalation, causing the Ubuntu command update-grub not to transition to the grub-mkconfig profile. This also has implications for zsys, which creates a new snapshot before every apt install and remove command and then adds the ZFS snapshot to the grub menu. The audit logs will be spammed with logging from zsysd because it doesn't transition. I've set zsysd hard-coded to complain mode.


apparmor.d/groups/avahi/avahi-browse
```diff
@@ -0,0 +1,32 @@
...
+      interface=org.freedesktop.Avahi.Server
+      member={GetAPIVersion,GetState,ServiceTypeBrowserNew,ServiceBrowserNew},
+
+  dbus receive bus=system path=/Client[0-9]/ServiceTypeBrowser[0-9]
+      interface=org.freedesktop.Avahi.ServiceTypeBrowser
+      member={ItemNew,CacheExhausted,AllForNow},
+
+  @{exec_path} rm,
+
+  /{usr/,}lib/x86_64-linux-gnu/avahi/service-types.db rwk,
```
roddhjav, line 29:
```suggestion
  /{usr/,}lib/@{multiarch}/avahi/service-types.db rwk,
```

apparmor.d/groups/freedesktop/xdg-open
```diff
@@ -10,6 +10,7 @@ include <tunables/global>
 profile xdg-open @{exec_path} flags=(attach_disconnected) {
   include <abstractions/base>
   include <abstractions/consoles>
+  include <abstractions/consoles>
```
roddhjav, line 13: Remove it, it is already there.

apparmor.d/groups/virt/k3s
```diff
@@ -26,7 +27,7 @@ profile k3s @{exec_path} {
   capability sys_resource,
 
   ptrace peer=@{profile_name},
-  ptrace (read) peer={cri-containerd.apparmor.d,cni-xtables-nft,kubernetes-pause,mount,unconfined},
```
roddhjav, line 29: Just figuring this out. What is `cri-containerd.apparmor.d` ?

apparmor.d/profiles-a-f/dkms
```diff
@@ -106,6 +110,9 @@ profile dkms @{exec_path} flags=(attach_disconnected) {
 
     /{usr/,}bin/kmod mr,
 
+    /etc/depmod.d/{,ubuntu.conf} r,
+    /etc/ssl/openssl.cnf r,
```
roddhjav, line 114: Use `include <abstractions/openssl>` instead.

apparmor.d/profiles-a-f/fwupdmgr
```diff
@@ -37,12 +60,12 @@ profile fwupdmgr @{exec_path} flags=(attach_disconnected,complain) {
   owner /var/cache/private/fwupdmgr/fwupd/lvfs-metadata.xml.gz{,.asc} rw,
 
   owner @{user_cache_dirs}/ rw,
+        @{user_cache_dirs}/dconf/user rw,
```
roddhjav, line 63: Remove, already present in `abstractions/dconf-write`

nobodysu:
Hi. Two cents:
Could you use `mr` instead of `rm`? This is confusing to newcomers.
Should we expose `/dev/pts` with `consoles` when only `/dev/tty` is needed, and wise versa? This also obscures single-line rule with general abstraction, which must be taken into account by the editor.

roddhjav:
> Could you use `mr` instead of `rm`? This is confusing to newcomers.

You are right, usually `mr` is used.

> Should we expose `/dev/pts` with consoles when only `/dev/tty` is needed, and wise versa? 

I do agree, exposing only `/dev/tty` is better than exposing `abstraction/consoles`.

roddhjav:
Also, `update-grub` should not be in the ubuntu group (as profile in this group are only present in ubuntu packages) whereas `update-grub` exist outside ubuntu

I am not sure it make sense to create profiles for tools like `whoami`, `whereis`, and `which` as these are tiny programs that are usually in `rix` mode in a profile. Also see this comment: https://github.com/roddhjav/apparmor.d/pull/11#issuecomment-1013729608

apparmor.d/groups/virt/k3s
Jeroen0494: Every container runtime has it's own default AppArmor profile. If you don't specify one (and don't run a privileged container) the default is applied. You can see the default template here: https://github.com/containerd/containerd/blob/main/contrib/apparmor/template.go#L43

Jeroen0494:
> > Could you use `mr` instead of `rm`? This is confusing to newcomers.
> 
> You are right, usually `mr` is used.

I can see the confusion. Okay, easy change.

> 
> > Should we expose `/dev/pts` with consoles when only `/dev/tty` is needed, and wise versa?
> 
> I do agree, exposing only `/dev/tty` is better than exposing `abstraction/consoles`.

I disagree. I access my server over SSH from my laptop, so the console is `/dev/pts/1`. My audit logs get spammed with these kind of AVC's all the time. Any and all user facing programs need access to `/dev/pts/` for me to read the output. The purpose of the `/dev/tty` is to print to console, so why not `/dev/pts`?

> Also, `update-grub` should not be in the ubuntu group (as profile in this group are only present in ubuntu packages) whereas `update-grub` exist outside ubuntu
> 

I wasn't sure because I run Ubuntu and currently nothing else, and I know it from Ubuntu and not Arch Linux, but I can move it.

> I am not sure it makes sense to create profiles for tools like `whoami`, `whereis`, and `which` as these are tiny programs that are usually in `rix` mode in a profile. Also see this comment: [#11 (comment)](https://github.com/roddhjav/apparmor.d/pull/11#issuecomment-1013729608)

For programs like `cp`, `mv` etc. absolutely. But `whoami`, `whereis`, and `which` are standalone programs, their access doesn't depend on the context of the program executing it. They don't need access to basically the entire filesystem like `cat` and `ls` do. The purpose of this project is listed as "this project tries to confine all the core applications you will usually find in a Linux system", and these standalone programs are part of every Linux system. There are profiles for other programs like these already, like `hostname`, `host`,  `man`. I guess you could argue the reason to confine these as well (I don't mean to come of as defensive btw, just explaining my thought process). I mean, it's your project, so you decide what you do and don't want to merge, but I figured if you want the core confined, these application are a part of it. Do you have a list of programs that still need to be confined / would rather see confined?

apparmor.d/profiles-a-f/fwupdmgr
Jeroen0494: Interesting, because it did show up in my audit log so it wasn't permitted. Will look into it.

apparmor.d/groups/freedesktop/xdg-open
Jeroen0494: I used a find and replace for all `/dev/tty` to consoles. I guess I missed a duplicate, will look for more.

roddhjav:
Actually, you are right, `which` is not `diff`, I was a bit fast to comment on this...

Interesting feedback on `consoles`. I have never had this kind of issues. So I think you can go ahead with it. Any thought @nobodysu?

nobodysu:
I saw `pts` noise occasionally, but didn't investigate because there was no actual failures. If it's needed - it's needed.

Jeroen0494:
I haven't tested every program yet, but every time I use a program I haven't before I get these `/dev/pts` AVC messages. So I figured, why not just use the consoles abstraction and prevent any AVC's before it becomes more of a problem.

apparmor.d/profiles-a-f/fwupdmgr
Jeroen0494: dconf-write has read only for `dconf/user`, this adds write.

Should write be part of the abstraction?

```
# Copyright (C) 2022 Alexandre Pujol <alexandre@pujol.io>

# Permissions for querying dconf settings with write access; use the dconf
# abstraction first, and dconf-write only for specific application's profile.

  /etc/dconf/** r,

  owner @{user_config_dirs}/dconf/user r,

  owner @{run}/user/@{uid}/dconf/ rw,
  owner @{run}/user/@{uid}/dconf/user rw,

  include if exists <abstractions/dconf-write.d>
```

nobodysu:
Since `consoles` brings wide exposure, I would refrain from bulk edits. Most likely, desktop applications doesn't need `pts`, and I would recheck whether `tty` is also needed.
Could you limit the edit to those profiles where you can actually see the `DENIED` entry?

Jeroen0494:
> 

I certainly can limit it to non-desktop application, if you require a more thorough scrub it will take more time. Perhaps it would be better in that case to revert commit https://github.com/roddhjav/apparmor.d/pull/64/commits/51a33f3f5e536b3d5c45e434e76339db3495e9cd and use a separate PR for this.

nobodysu:
The final decision isn't mine, I'm just expressing an opinion.
I think `consoles` should be limited to user-callable programs from the shell, meaning utilities and some services.

apparmor.d/profiles-s-z/udisksd
```diff
@@ -130,8 +130,7 @@ profile udisksd @{exec_path} flags=(attach_disconnected) {
   @{sys}/class/ r,
 
   @{sys}/devices/pci[0-9]*/**/{ata,usb,mmc}[0-9]/{,**/}uevent w,
-  @{sys}/devices/virtual/block/dm-[0-9]*/ w,
-  @{sys}/devices/virtual/block/dm-[0-9]*/** w,
+  @{sys}/devices/virtual/block/*/** r,
```
roddhjav, line 133: Any logic behind this?
1. Please note this profile has the disks-write abstraction already.
2. Beware with changes in `udisksd` regarding disks, that the best way to break some system.

Jeroen0494: With `mdadm` you can give your raid array any name you like, and udiskd has to write to it, so only` dm-[0-9]*` is not enough. I think the read permission without write is a typo, I'll fix that.

roddhjav:
> I think consoles should be limited to user-callable programs from the shell, meaning utilities and some services.

I do agree on this one.

apparmor.d/profiles-a-f/fwupdmgr
roddhjav: First time I see this need. So let's keep it there for now.

roddhjav:
> I certainly can limit it to non-desktop application, if you require a more thorough scrub it will take more time. Perhaps it would be better in that case to revert commit https://github.com/roddhjav/apparmor.d/commit/51a33f3f5e536b3d5c45e434e76339db3495e9cd and use a separate PR for this.

That would be the best option actually. Revet the commit, so I can merge this PR and it give time for a new PR specific for consoles matter.

roddhjav:
Perfect. Merged!

PR #65, opened 2022-08-18
Ubuntu 22.04, third batch

nobodysu:


apparmor.d/profiles-a-f/engrampa
```diff
@@ -69,9 +111,18 @@ profile engrampa @{exec_path} {
...
 
+  /etc/magic r,
+
   /usr/share/glib-2.0/schemas/gschemas.compiled r,
 
-  /etc/magic r,
+  # gnome-tiny
+  @{run}/mount/utab r,
+
+  # Ubuntu
```
nobodysu, line 121: Any suggestions for an abstraction?

apparmor.d/groups/browsers/firefox
```diff
@@ -9,7 +9,8 @@ abi <abi/3.0>,
 
 include <tunables/global>
 
-@{MOZ_LIBDIR} = /{usr/,}lib/firefox{,-esr}
+@{MOZ_LIBDIR}  = /{usr/,}lib/firefox{,-esr}
+@{MOZ_LIBDIR} += /opt/firefox{,-esr}
```
roddhjav, line 13: Out of curiosity, how did you install Firefox to get it in /opt ?

apparmor.d/groups/browsers/firefox
```diff
@@ -154,7 +233,6 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
        @{sys}/devices/pci[0-9]*/**/drm/renderD[0-9]*/ r,
        @{sys}/devices/pci[0-9]*/**/irq r,
        @{sys}/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us r,
-       @{sys}/devices/system/cpu/possible r,
```
roddhjav, line 157: Why?

apparmor.d/profiles-a-f/engrampa
roddhjav: Logically the applications resources should go into `abstractions/freedesktop.org`

apparmor.d/groups/browsers/firefox
nobodysu: Oops, sorry, outdated master on my end.

apparmor.d/groups/browsers/firefox
nobodysu: Manual installation from official tar. It's a natural place for third-party applications and present as example on many FF guides.

nobodysu:
Started to write dbus rules without parenthesis and omitted an action by mistake (still worked, but that means all actions). IMO, ditching control syntax is a source of errors. We still have an established `unix (...)` and `signal (...)` style btw.

roddhjav:
Don't worry too much about parenthesis, do like you feel the best is.

One day, I will write a linter and a formatter to help us working with profile. That will definitely solve this question.

roddhjav:
Thanks. Merged!

PR #70, opened 2022-09-03
Remove `lightdm` abstractions

nobodysu:
Also I suggest to make Debian/Ubuntu all-complain by default, because right now the package just breaks the system.
And, setting `--complain` from command line instead of editing the file before building will be very convenient.


README.md
```diff
@@ -63,13 +63,14 @@ sudo pacman -U apparmor.d-*.pkg.tar.zst \
...
 > **Warning**: for a first install, it is recommanded to install all profiles in complain mode. See [Complain mode](#troubleshooting)
 
-**Debian**
+**Debian / Ubuntu**
 
 Build using standard Debian package build tools:
 ```sh
-sudo apt install apparmor-profiles build-essential config-package-dev debhelper golang-go rsync
+sudo apt install apparmor-profiles build-essential config-package-dev debhelper golang-go rsync git
+git clone https://github.com/roddhjav/apparmor.d.git && cd apparmor.d
```
nobodysu, line 71: This is not very obvious to an average user.

README.md
```diff
@@ -63,13 +63,14 @@ sudo pacman -U apparmor.d-*.pkg.tar.zst \
...
 dpkg-buildpackage -b -d --no-sign
-sudo dpkg -i ../apparmor.d_*_all.deb
+sudo dpkg --force overwrite -i ../apparmor.d_*_all.deb
```
nobodysu, line 73:
```
dpkg: error processing archive ../apparmor.d_0.001-1_all.deb (--install):
 trying to overwrite '/etc/apparmor.d/abstractions/lightdm', which is also in package lightdm 1.26.0-7
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 ../apparmor.d_0.001-1_all.deb
```

roddhjav: This is not the correct way of doing this. You should add `/etc/apparmor.d/abstractions/lightdm.apparmor.d` to `debian/apparmor.d.displace`

roddhjav:
The `--complain` option is designed for dev only. So yes as the full project is still in dev stage we could put the default to complain mode and then tell the user to remove it only after testing it. So quite the opposite of what we do actually.

However, even during the dev stage it should not block the system. To prevent this, I setup a system to easily set flags on work in progress profile in `dists/flags/main.flags`.  All profiles that may break the system (or non production ready profile) should have an entry in this file. This, it allow us to enforce all the other profiles that are known to work well.

README.md
nobodysu: Hmm, that would be the same error.

nobodysu:
```
To prevent this, I setup a system to easily set flags on work in progress profile in dists/flags/main.flags. All profiles that may break the system (or non production ready profile) should have an entry in this file. This, it allow us to enforce all the other profiles that are known to work well.
```
Understood. I do not have the list of fatal profiles at the time.

README.md
roddhjav: It should work fine. This is how `/etc/apparmor.d/tunables/xdg-user-dirs` is displaced.

Now, in this context, the solution is to remove `abstractions/lightdm` and `apparmor.d/abstractions/lightdm_chromium-browser` from this project are we are not modifying them. These files were added by morfikov, but since my fork evolved, we don't overwrite other package profiles unless it is strictly required.

nobodysu:
I'll test it more thoroughly.

nobodysu:
Seems to work.

roddhjav:
Thanks. Merged!

PR #78, opened 2022-09-23
viewing DjVu and PostScript files

beroal:


apparmor.d/profiles-a-f/evince
```diff
@@ -24,12 +24,15 @@ profile evince @{exec_path} {
...
 
+  /usr/share/djvu/{,**} r,
   /usr/share/evince/{,**} r,
+  /usr/share/ghostscript/{,**} r,
   /usr/share/poppler/{,**} r,
   /usr/share/thumbnailers/{,*} r,
   /usr/share/themes/{,**} r,
 
   owner @{user_share_dirs}/ r,
+  owner @{user_share_dirs}/gvfs-metadata/{,*} r,
```
nobodysu, line 35: Could this be denied? This is a major information leak and evince is purposed for deliberately risky files.

roddhjav: I think it can.

roddhjav:
Merged thanks!

I will deny the `gvfs-metadata`path later (I am looking at denying everywhere it is not strictly required).

PR #81, opened 2022-10-05
Unbreak Debian 11 and partially Ubuntu 22.04 (Wayland+GDM+Gnome)

nobodysu:
Almost nothing deleted, but due to adding peer's labels it *may* break some DE combinations (besides Wayland+GDM+Gnome). The path to a fully working DE ruleset is seen to me as number of iterations when errors appear less and less frequently. Due to DBus complexity it may be tool-assisted. (No ETA on both)
Once DBus rules will crystalize, they may go into several abstractions.

Opinion:
- peer's labels should be mandatory, except when there's no peers or they are too broad
- for transitioning reasons, when peer's label is absent it should be explicitly stated why (no peers or too broad)
- at some point we should ditch integrated DBus abstractions in favor of stricter label-mentioned ones
- current collapse policy of DBus rules is too aggressive, honestly, I can't collapse them any more than this, it's mandatory access after all


apparmor.d/groups/bus/ibus-x11
```diff
@@ -21,12 +22,34 @@ profile ibus-x11 @{exec_path} flags=(attach_disconnected) {
...
+       peer=(name=org.a11y.atspi.Registry), # all peer's labels
+
   @{exec_path} mr,
 
-  /var/lib/gdm{3,}/.config/ibus/bus/{,@{hex}-unix-wayland-[0-9]} r,
+  /var/lib/gdm{3,}/.config/ibus/bus/ r,
+  /var/lib/gdm{3,}/.config/ibus/bus/@{hex}-unix{,-wayland}-[0-9] r,
+  /var/lib/gdm{3,}/.cache/mesa_shader_cache/index rw,
 
-  owner @{user_config_dirs}/ibus/bus/{,@{hex}-unix-wayland-[0-9]} r,
```
nobodysu, line 28: This approach is too obscure.

apparmor.d/groups/systemd/systemd-timesyncd
```diff
@@ -23,15 +23,15 @@ profile systemd-timesyncd @{exec_path} flags=(attach_disconnected) {
...
+       member={RequestName,ReleaseName}
+       peer=(name=org.freedesktop.DBus, label=dbus-daemon),
 
   dbus bind bus=system
        name=org.freedesktop.timesync1,
 
   @{exec_path} mr,
 
-  /etc/adjtime r,
+  @{etc_rw}/adjtime r,
```
nobodysu, line 34: Readonly root.

roddhjav:
Thanks for your PR!! Nice work on the dbus session rule. It has been a while I wanted to add these rules.

The main issue I see with your current approach is that it is way to complex to maintain. I mean look at `gnome-shell` (Furthermore, as the main DE profile it might be a special case and just allowing it `dbus` may be a better solution.)

My idea regarding DBus complexity issue was to create a (quite large) set of
abstraction allowing us to easily add access to some resource in similar way that sandbox manager like Flatpak does.

For example Flatpak has this structure:
* Talks: Allow the application to talk to session services. (send)
* Owns: Allow the application to own session services under the given name. (receive, send, bind)

Furthermore it has preconfigured access for some defined portal: `org.freedesktop.portal.*`
See: https://github.com/tchx84/Flatseal/blob/master/DOCUMENTATION.md#session-bus

So we could have similar construction and get abstraction like  `dbus-camera`, `dbus-notification`, `dbus-microphone`, `dbus-speakers`, `dbus-location`...

> peer's labels should be mandatory, except when there's no peers or they are too broad

Might be hard to maintain and to organize the abstraction with it, but we can try it sure.

> at some point we should ditch integrated DBus abstractions in favor of stricter label-mentioned ones

Are you sure about this. I mean, doing this will result in very very long profile full of unreadable dbus rule. Please let's keep these profile readable and manageable by an human.

> current collapse policy of DBus rules is too aggressive, honestly, I can't collapse them any more than this, it's mandatory access after all

I am open to suggestion here. As I am not a dbus expert I originally kept all dbus request as it, so with all `interface`, `member`, `path`... This is too strict, we need to revisit this way of writing the rules. For example, I don't know if restricting some `member` under a dbus path make sense. Possibly we should limit ourselves to something that: `own`, `talks` over a given interface. That would be way easier to maintain as it would not require adding new rule on every update of the profiled program.

nobodysu:
>The main issue I see with your current approach is that it is way to complex to maintain. I mean look at gnome-shell

We'll maintain system profiles once every 4 years and give broader access to application-related rules. This is an attempt to see if it's even possible to cover profile with such a large scope with strict rules. If it won't work out - we'll just assign whole `dbus` abstraction.

>So we could have similar construction and get abstraction like dbus-camera, dbus-notification, dbus-microphone, dbus-speakers, dbus-location...

I had a similar idea. This PR is a transition before we could see the whole picture and compose these abstractions. It dbus rules are collapsed aggressively right now we would not be able to see it.

>Might be hard to maintain and to organize the abstraction with it, but we can try it sure.

I think this is intended by design. Most of the rules have only one peer's label (caller/callee). If access is broad - we'll just omit the label.

>Are you sure about this. I mean, doing this will result in very very long profile full of unreadable dbus rule. Please let's keep these profile readable and manageable by an human.

This is trivial. I'm talking about this:
```
  dbus send bus=system path=/org/freedesktop/DBus
       interface=org.freedesktop.DBus
       member={Hello,AddMatch,RemoveMatch,GetNameOwner,NameHasOwner,StartServiceByName}
       peer=(name=org.freedesktop.DBus, label=dbus-daemon),
```
I haven't seen any other callee for `/org/freedesktop/DBus`. Upstream doesn't have it because it lacks profile for `dbus-daemon`.

>I am open to suggestion here. As I am not a dbus expert I originally kept all dbus request as it, so with all interface, member, path... This is too strict, we need to revisit this way of writing the rules.

Me neither. I can only see the collapse in suffixes, otherwise it's just different rules with different caller/callee. `Members` are excessive for services that holds them. Again, peer's labels will help very much with narrowing the access, e.g.:
```
  dbus receive bus=system path=/org/freedesktop/login[0-9]*
       interface=org.freedesktop.login[0-9]*.Manager
       peer=(name=:*, label=systemd-logind),
```
>That would be way easier to maintain as it would not require adding new rule on every update of the profiled program.

DBus is so complex that tool is needed. I'm designing it but there's no ETA and it might not work out.
Right now I'm suggesting to go towards forming the abstractions.

roddhjav:
Good. It seems we have the same vision on this subject. So let's move forward, test and break things to improve our dbus integration. Tell me when you think the PR can be merged.

> DBus is so complex that tool is needed. I'm designing it but there's no ETA and it might not work out.
Right now I'm suggesting to go towards forming the abstractions.

What do you mean by tool. To do what? Automate dbus rule gen?

nobodysu:
>What do you mean by tool. To do what? Automate dbus rule gen?

I see it as visual representation of caller/callee, with `members` and `path`'s suffixes grouped. Rule generation might be a side effect.

>Tell me when you think the PR can be merged.

I'll polish the PR and undraft it in following days.

apparmor.d/groups/freedesktop/pipewire
```diff
@@ -27,6 +32,11 @@ profile pipewire @{exec_path} {
        member=Get
        peer=(name=org.freedesktop.RealtimeKit[0-9]),
 
+  dbus receive bus=session path=/
+       interface=org.freedesktop.DBus.Introspectable
+       member=Introspect
+       peer=(name=:*, label=gnome-shell),
```
nobodysu, line 38: Likely candidate as `dbus-gnome-subject`.

apparmor.d/groups/bus/ibus-portal
```diff
@@ -29,7 +41,6 @@ profile ibus-portal @{exec_path} flags=(attach_disconnected) {
   /var/lib/gdm{3,}/.config/ibus/bus/@{hex}-unix-{,wayland-}[0-9] r,
 
   owner /dev/tty[0-9]* rw,
-  /dev/null rw,
```
nobodysu, line 32: Abstraction.

dists/flags/ubuntu.flags
```diff
@@ -11,6 +11,7 @@ notify-reboot-required complain
 package-system-locked attach_disconnected,complain
 packagekitd complain
 release-upgrade-motd complain
+setpriv complain
```
nobodysu, line 14: I wonder how Arch behaves without this.

apparmor.d/groups/gnome/gnome-shell
```diff
@@ -504,7 +505,6 @@ profile gnome-shell @{exec_path} flags=(attach_disconnected) {
   /etc/fstab r,
   /etc/xdg/menus/gnome-applications.menu r,
 
-  /var/lib/gdm{3,}/.cache/ w,
```
roddhjav, line 88: Put it back. It is needed to create the directory on first start.

apparmor.d/groups/gvfs/gvfsd-metadata
```diff
@@ -3,7 +3,6 @@
 # Copyright (C) 2022 Alexandre Pujol <alexandre@pujol.io>
 # SPDX-License-Identifier: GPL-2.0-only
 
-abi <abi/3.0>,
```
roddhjav, line 6: Put it back.

apparmor.d/groups/systemd/systemd-hostnamed
```diff
@@ -8,7 +8,7 @@ abi <abi/3.0>,
 include <tunables/global>
 
 @{exec_path} = /{usr/,}lib/systemd/systemd-hostnamed
-profile systemd-hostnamed @{exec_path} flags=(attach_disconnected) {
+profile systemd-hostnamed @{exec_path}  flags=(attach_disconnected complain) {
```
roddhjav: Do add `complain` here. Add it in the `dists/flags/main.flag` file.

apparmor.d/profiles-a-f/cc-remote-login-helper
```diff
@@ -0,0 +1,17 @@
+# apparmor.d - Full set of apparmor profiles
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{libexec}/cc-remote-login-helper
+profile cc-remote-login-helper @{exec_path} flags=(complain) {
```
roddhjav, line 9: Do not add `complain` here. Add it in the `dists/flags/main.flag` file.

apparmor.d/groups/systemd/systemd-hostnamed
nobodysu: It's fixed, that's not a final commit.

apparmor.d/profiles-a-f/cc-remote-login-helper
nobodysu: Are any flags discouraged or just complain?

apparmor.d/groups/gvfs/gvfsd-metadata
nobodysu: Intermediary commit. It's fixed in the latest.

apparmor.d/groups/gnome/gnome-shell
nobodysu: Sorry, mistook for my own change. There are many situations like this, I postponed them for a future PR.

nobodysu:
- Debian 11 with Wayland+GDM+Gnome: boot, login, networking, control-center fully working. Userland apps are not tested.
- Ubuntu 22.04 with Wayland+GDM+Gnome: boot, login, networking, control-center partially working, still many DBus errors. Userland apps are not tested.

Hovewer, I haven't been able to test `full` mode:
```
apparmor.d# ./configure --full
Set the configuration for ubuntu.
 - Ignore profiles/files in dists/ignore/main.ignore
 - Ignore profiles/files in dists/ignore/ubuntu.ignore
 - Synchronise all profiles.
 - Configure libexec.
 - Displace overwritten files.
 - Set profiles flags from dists/flags/main.flags
   Warning: Profile docker not found
   Warning: Profile man not found
 - Set profiles flags from dists/flags/ubuntu.flags
cp: cannot stat 'apparmor.d/groups/_full/*': No such file or directory
```

Future plans:
- Mitigations for new user login
- full peer's label coverage, when applicable
- Ungrouping of DBus rules, when applicable

apparmor.d/profiles-a-f/cc-remote-login-helper
roddhjav: Just complain. As it should be removed at some point,  I do it this way as it is easier to have a central place to maintain the flags for all unstable profiles.

dists/flags/ubuntu.flags
roddhjav: I have never seen it on arch. Anyway, as it is not ubuntu/debian specific, you should move it in `dists/flags/main.flags`

roddhjav:
> * Ubuntu 22.04 with Wayland+GDM+Gnome: boot, login, networking, control-center partially working, still many DBus errors. Userland apps are not tested.

The difficulty here is that Ubuntu adds a lot of internal software (plus snap) that is pain to integrate. Plus there is the dbus subject as Ubuntu is the only distribution that has the dbus integration enabled. 

The full mode is not ready yet, (not even committed yet), I only added support in the `configure` script.

> Read-only root compatible

That's part of my plan too. I want to test this project with a stateless system with readonly root image generated with mkosi. As it is the next big evolution in Linux distribution and as it has a huge security impact, it seems logic test it with apparmor.

apparmor.d/groups/bus/dbus-daemon
```diff
@@ -47,6 +48,7 @@ profile dbus-daemon @{exec_path} flags=(attach_disconnected) {
   /{usr/,}lib/@{multiarch}/tumbler-1/tumblerd    rPUx,
 
   /usr/share/org.gnome.Characters/org.gnome.Characters.BackgroundService rPx,
+  /usr/share/gnome-documents/org.gnome.Documents rPx,
```
roddhjav, line 51: Where is the profile for this?

apparmor.d/groups/bus/dbus-daemon-launch-helper
```diff
@@ -22,8 +22,11 @@ profile dbus-daemon-launch-helper @{exec_path} {
   /{usr/,}lib/cups-pk-helper-mechanism                      rPx,
   /{usr/,}lib/cups-pk-helper/cups-pk-helper-mechanism       rPx,
   /{usr/,}lib/software-properties/software-properties-dbus  rPx,
+  @{libexec}/language-selector/ls-dbus-backend              rPx,
```
roddhjav, line 25: Same, no profile for this. So either you add a profile or you `rPUx` as a temporary solution before adding the profile.

apparmor.d/groups/bus/dbus-daemon-launch-helper
```diff
@@ -22,8 +22,11 @@ profile dbus-daemon-launch-helper @{exec_path} {
...
 
   /usr/share/org.gnome.Characters/org.gnome.Characters.BackgroundService rPx,
+  /usr/share/usb-creator/usb-creator-helper rPx,
+  /usr/share/hplip/pkservice.py rPx,
```
roddhjav, line 29: Same here, no profile.

apparmor.d/groups/bus/dbus-daemon
nobodysu: Nowhere yet. I felt reluctant setting `rPUx` for minor processes from major ones (`dbus-daemon`, `pkexec`). There no fatal errors. You still want `rPUx`?

roddhjav:
Merged! I will add the `rPUx` myself when needed.

apparmor.d/groups/bus/dbus-daemon
roddhjav: Yes, otherwise this is a bug at run time.

PR #91, opened 2022-12-02
Update CONTRIBUTING.md

nobody43:
For refactoring purposes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,6 +114,7 @@ The rules in the profile should be sorted in rule *block* as follow:
 - unix
 - dbus
 - file
+- file_inherit
 - Local include
 
 This rule order is taken from AppArmor with minor changes as we tend to:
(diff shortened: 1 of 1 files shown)


roddhjav:
Hi, I am not sure I agree with this rule.

For `file_inherit`, it modesty depends from witch software it is inheriting file descriptor. As a lot of profile works in collaboration with other this lead having a lot of perfectly legitimate `file_inherit`.

> Keeping profiles up to the current guideline is encouraged but not mandatory.

Yes, but only because it is not possible to update all of them (and because the guideline itself is quite young). However, all new profile must follow it. So not sure this wording make sense here.

nobody43:
As I understand, situation with [fd delegation will change at some point](https://gitlab.com/apparmor/apparmor/-/issues/92). And then we will end up with stale unnecessary entries. Maybe it's good to keep track of them from the start?

> However, all new profile must follow it. So not sure this wording make sense here.

Judging from my experience, I was slightly confused by different styles in profiles and `CONTRIBUTING.md`.

nobody43:
Alright, lets think about it some more. I'm sure we'll return to this matter.

PR #93, opened 2023-01-06
Add profiles

nobody43:


apparmor.d/profiles-g-l/lsd
```diff
@@ -0,0 +1,22 @@
...
+  include <abstractions/base>
+  include <abstractions/nameservice-strict>
+  include <abstractions/consoles>
+
+  capability dac_read_search,
+
+  @{exec_path} mr,
+
+  / r,
+  /**/ r,
```
nobody43, line 19: Thoughts? Third-party, sure we don't want to spin up a sandbox for directory traverse.

apparmor.d/profiles-m-r/rustdesk
```diff
@@ -0,0 +1,143 @@
+# apparmor.d - Full set of apparmor profiles
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = /{,usr/}{,local/}bin/rustdesk
```
nobody43, line 8: Why `local`? Because it's a bleeding edge software which is absent in the package repository.

apparmor.d/profiles-a-f/btop
```diff
@@ -0,0 +1,49 @@
...
+
+  network netlink raw,
+
+  signal (send),
+  ptrace (read),
+
+  @{exec_path} mr,
+
+  /etc/fstab r,
+  /etc/passwd r,
```
roddhjav, line 23: Remove, and use `include <abstractions/nameservice-strict>` instead.

apparmor.d/profiles-a-f/btop
```diff
@@ -0,0 +1,49 @@
...
+  @{sys}/devices/{pci*,virtual}/{,*/*/}net/*/address r,
+  @{sys}/devices/pci*/*/*/usb3/*/*/*/*/power_supply/hidpp_battery_[0-9]*/{,hwmon[0-9]*/} r,
+
+        @{PROC} r,
+        @{PROC}/loadavg r,
+        @{PROC}/uptime r,
+        @{PROC}/@{pid}/comm r,
+        @{PROC}/@{pid}/cmdline r,
+        @{PROC}/@{pid}/stat r,
+        @{PROC}/@{pid}/io r,
```
roddhjav, line 45: For a better result with future version of AA. Use `@{pids}` here are these can be any process id, not only the one confined by this profile.

apparmor.d/profiles-g-l/lsd
```diff
@@ -0,0 +1,22 @@
+# apparmor.d - Full set of apparmor profiles
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = /{,usr/}{,local/}bin/lsd
```
roddhjav, line 8: What is lsd? If it is this [new ls](https://github.com/Peltoche/lsd), you should not confine it at all.

apparmor.d/profiles-m-r/murmurd
```diff
@@ -0,0 +1,53 @@
...
+  network inet stream,
+  network inet6 stream,
+  network inet dgram,
+  network inet6 dgram,
+
+  @{exec_path} mr,
+
+  /{,usr/}bin/lsb_release Px -> lsb_release,
+
+  dbus send bus=system path=/
```
roddhjav, line 29: Try to put the `dbus` rules before the `@{exec_path} mr,`

apparmor.d/profiles-m-r/murmurd
```diff
@@ -0,0 +1,53 @@
...
+  owner @{run}/mumble-server/*.pid rw,
+
+  # Silencer
+  deny / r,
+  deny /usr/{,local/}lib/ r,
+  deny /usr/lib32/ r,
+  deny /usr/lib64/ r,
+
+  # file_inherit
+  unix (send, receive) type=stream addr=none peer=(label=lsb_release),
```
roddhjav, line 50: That should go before the `@{exec_path} mr,` too.

apparmor.d/profiles-m-r/rustdesk
```diff
@@ -0,0 +1,143 @@
...
+       interface=org.freedesktop.DBus.Properties
+       member=Set
+       peer=(name=:*),
+
+  owner /tmp/[rR]ust[dD]esk/{,**} rw,
+
+  owner @{user_share_dirs}/logs/[rR]ust[dD]esk/{,**} rw,
+  owner @{user_config_dirs}/[rR]ust[dD]esk/{,**} rw,
+
+  owner @{run}/user/@{uid}/at-spi/bus{,_[0-9]*} rw,
```
roddhjav, line 61: This is already part of `abstractions/dbus-session-strict`

apparmor.d/profiles-m-r/rustdesk
```diff
@@ -0,0 +1,143 @@
...
+#  ptrace (read),
+#  owner @{PROC}/@{pid}/stat r,
+#  owner @{PROC}/@{pid}/environ r,
+#  owner @{PROC}/@{pid}/io r,
+#  owner @{PROC}/@{pid}/task/ r,
+#  owner @{PROC}/@{pid}/task/@{tid}/stat r,
+#  owner @{PROC}/@{pid}/task/@{tid}/io r,
+#  owner @{PROC}/@{pid}/task/@{tid}/status r,
+
+  profile python {
```
roddhjav, line 102: Nice idea to put the python part of a program done in rust in its own subprofile.
Do you know the purpose of this python segment? 

These lines make me think that could be they update system:
```
   /usr/share/dpkg/cputable r,
    deny /etc/apt/{,**} r,
```

In this case `/etc/apt/` might be needed

nobody43: >Do you know the purpose of this python segment?

This is somehow related to file transfer, and not essential.

>These lines make me think that could be they update system:

Then we shouldn't allow this access to a remote control program. I'm thinking it's some python artifact, whether related to Xubuntu or my specific installation, I'm not sure yet.

apparmor.d/profiles-m-r/murmurd
nobody43: Sorry, got confused with recent style change.

apparmor.d/profiles-g-l/lsd
nobody43: It's a third-party program, why it should be trusted?

apparmor.d/profiles-a-f/btop
```diff
@@ -0,0 +1,49 @@
...
+  @{sys}/class/hwmon/ r,
+  @{sys}/devices/system/cpu/cpufreq/policy[0-9]*/scaling_{cur,min,max}_freq r,
+  @{sys}/devices/virtual/thermal/thermal_zone[0-9]*/ r,
+  @{sys}/devices/virtual/thermal/thermal_zone[0-9]*/hwmon[0-9]*/{,*} r,
+  @{sys}/devices/platform/coretemp.[0-9]*/hwmon/hwmon[0-9]*/{,*} r,
+  @{sys}/devices/pci*/*/*/host[0-9]*/*/*/block/*/*/stat r,
+  @{sys}/devices/virtual/block/dm-[0-9]*/stat r,
+  @{sys}/devices/{pci*,virtual}/{,**/}net/*/statistics/{rx,tx}_bytes r,
+  @{sys}/devices/{pci*,virtual}/{,*/*/}net/*/address r,
+  @{sys}/devices/pci*/*/*/usb3/*/*/*/*/power_supply/hidpp_battery_[0-9]*/{,hwmon[0-9]*/} r,
```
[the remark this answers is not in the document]
nobody43: >Shouldn't this be similar to the htop profile?

It's not asking for anything additional.

apparmor.d/profiles-m-r/rustdesk
roddhjav: That is not python related, this is dpkg resources (See the apt-common abstraction).

> This is somehow related to file transfer, and not essential.

Ok, let's keep it like this for now, might need more resources later (like network access)

apparmor.d/profiles-g-l/lsd
nobody43: Yeah, that first attempt was bad, but how custom `ls` could break the system? It's not replacing `ls` - it won't break scripts. Pipes? User could modify `local` drop-in for edge cases. If we'll account for pipes and edge cases half of profiles would be broken...

roddhjav:
Sorry for the delay with this PR. Once you remove the `lsd` profile, it should be good to go.

roddhjav:
Thanks a lot. Merged!

PR #96, opened 2023-01-13
General update

Jeroen0494:
Several updates discovered after a system update.

`/usr/share/debconf/frontend` executes `update-secureboot-policy`, but `frontend` has no policy so there are still some AVC messages left.


roddhjav:
Thanks a lot. Merged!

PR #106, opened 2023-01-21
add support for cachyos

name-tar-xz:
Adds support for building on Arch based distro [CachyOS](https://cachyos.org/)


roddhjav:
Thanks a lot. Merged!

PR #114, opened 2023-02-01
gpgconf: read @{run}/user/@{uid}/gnupg/

vbauerster:
use case:

```
ALLOWED gpgconf mkdir /run/user/1000/gnupg/ comm=gpgconf requested_mask=c denied_mask=c
```


roddhjav:
That should be write access, not read?

vbauerster:
yes I missed `c` mask. Does it mean write?

roddhjav:
Well it means create, which means the rule need the `w`rite access.

PR #120, opened 2023-02-19
General update

nobody43:


apparmor.d/profiles-m-r/mount
```diff
@@ -9,7 +9,7 @@ abi <abi/3.0>,
 include <tunables/global>
 
 @{exec_path} = /{usr/,}{s,}bin/mount
-profile mount @{exec_path} flags=(complain) {
```
nobody43, line 12: Already handled in `main.flags`?

apparmor.d/groups/freedesktop/pulseaudio
```diff
@@ -113,10 +114,8 @@ profile pulseaudio @{exec_path} {
   /{usr/,}lib/@{multiarch}/pulse/gconf-helper mrix,
   /{usr/,}lib/pulse-*/modules/*.so mr,
 
-  /usr/share/applications/{,**} r,
   /usr/share/glib-2.0/schemas/gschemas.compiled r,
```
nobody43, line 117: Add to `abstractions/freedesktop.org.d`?

apparmor.d/profiles-m-r/qbittorrent
```diff
@@ -31,11 +31,13 @@ profile qbittorrent @{exec_path} {
   include <abstractions/dbus-accessibility-strict>
   include <abstractions/dbus-network-manager-strict>
   include <abstractions/dbus-gtk>
+  include <abstractions/dri-common>
   include <abstractions/dri-enumerate>
   include <abstractions/wayland>
   include <abstractions/mesa>
   include <abstractions/openssl>
   include <abstractions/ssl_certs>
+  include <abstractions/nvidia>
```
nobody43, line 40: We should start to think about something like `gpu` abstraction.

apparmor.d/groups/virt/virtiofsd
```diff
@@ -0,0 +1,47 @@
+# apparmor.d - Full set of apparmor profiles
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{LOCAL_SHARED_DIRS} = /var/lib/libvirt/shared
+include if exists <local/tunables/virtiofsd>
```
nobody43, line 9: This opens up the possibility to adjust local variables. Thoughts?

apparmor.d/groups/freedesktop/pulseaudio
roddhjav: Yes, I find that could be a good option.

apparmor.d/profiles-m-r/qbittorrent
roddhjav: `abstractions/opencl` is pretty mucj a gpu abstraction already.

apparmor.d/groups/virt/virtiofsd
roddhjav: Adjust by the user? no. For this you need to use [global variables](https://apparmor.pujol.io/variables/). Variables defined in a profile header are not modifiable outside.

Here, you should have something like I did in `libvirtd`: https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/groups/virt/libvirtd#L146-L150

apparmor.d/groups/browsers/firefox
```diff
@@ -12,7 +12,7 @@ include <tunables/global>
 @{firefox_config_dirs} = @{HOME}/.mozilla/
 @{firefox_cache_dirs} = @{user_cache_dirs}/mozilla/
 
-@{exec_path} = /{usr/,}bin/@{firefox_name} @{firefox_lib_dirs}/@{firefox_name}
+@{exec_path} = /{usr/,}bin/@{firefox_name} @{firefox_lib_dirs}/@{firefox_name}{-bin,}
```
roddhjav, line 15: Edit `@{firefox_name}` instead

apparmor.d/groups/browsers/firefox
```diff
@@ -173,6 +173,8 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
   /etc/xul-ext/kwallet5.js r,
 
   owner @{HOME}/ r,
+  owner @{HOME}/@{XDG_DESKTOP_DIR}/ w,
+  owner @{HOME}/@{XDG_DOWNLOAD_DIR}/ w,
```
roddhjav, line 177: I think you should edit `abstractions/user-download-strict` instead.

apparmor.d/profiles-m-r/mount
roddhjav: They should. But when removing complain flags here please double check that they are already in `main.flags`.

roddhjav: Also, you should add all your new profile in `main.flags`.

apparmor.d/profiles-m-r/remmina
```diff
@@ -0,0 +1,145 @@
...
+  owner @{PROC}/@{pid}/task/@{tid}/comm rw,
+  owner @{PROC}/@{pid}/mountinfo r,
+
+  owner @{run}/user/@{uid}/keyring/ssh rw,
+
+  # gtk-tiny
+  /etc/gtk-3.0/settings.ini r,
+  /usr/share/themes/{,**} r,
+
+  # X-strict
```
roddhjav, line 138: Why don't you use the X-strict abstraction?

roddhjav:
Thanks a lot. Nice work with the dbus rules!

Regarding `loginctl`, move it to the systemd group. Also you need to add:
```
  /{usr/,}bin/less  rPx -> child-pager,
  /{usr/,}bin/more  rPx -> child-pager,
  /{usr/,}bin/pager rPx -> child-pager,
```

As well as `include <abstractions/systemd-common>` in it (this profile is part of some un-published profile I have so I did a few test with it already)

Regarding `virtiofsd`, I have already played with it a lot and I have never found proper way to profile it.

apparmor.d/profiles-m-r/mount
nobody43: Double checked. Added.

apparmor.d/groups/freedesktop/pulseaudio
nobody43: Wow, I'll move it in a separate PR:
```
grep -ri '/usr/share/glib-2.0/schemas/gschemas.compiled' apparmor.d | wc -l
131
```

apparmor.d/profiles-m-r/qbittorrent
nobody43: Close, but to broad for my liking.

apparmor.d/profiles-m-r/remmina
nobody43: That's an intermediary for the following xorg-related PR.

nobody43:
Note: I haven't been able to test `virtiofsd` with project's `libvirtd` yet, so it might need it's following `@{user_vm_shares}` rule.

nobody43:
I'm not a fan of whitespaces in Linux, but that's what distros already do, changed.
Not dotfile because the directory is intended for GUI access.
I'm trying not to give gross-access unless needed.

roddhjav:
Look good. Merged thanks.

PR #125, opened 2023-02-23
General update

Jeroen0494:
Hi!

Another batch of general updates all over. Some dbus fixes for PulseAudio, some fixes for apt when using automatic updates, some containerd fixes.

### DIFF
--- a/apparmor.d/groups/apt/apt-methods-gpgv
+++ b/apparmor.d/groups/apt/apt-methods-gpgv
@@ -69,11 +69,11 @@ profile apt-methods-gpgv @{exec_path} {
   /etc/ r,
   /root/ r,
 
-        /var/lib/apt/lists/{,**} r,
+        {/var/lib/ubuntu-advantage/apt-esm,}/var/lib/apt/lists/{,**} r,
         /var/lib/dpkg/arch r,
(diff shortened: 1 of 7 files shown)


Jeroen0494:
Hmm, interesting. I get a systemctl AVC with apt, but the line already exists in the profile.
```
ALLOWED apt exec /usr/bin/id comm=sh requested_mask=x denied_mask=x
ALLOWED apt exec /usr/bin/systemctl comm=sh requested_mask=x denied_mask=x
```

apparmor.d/groups/apt/apt-methods-gpgv
```diff
@@ -69,11 +69,11 @@ profile apt-methods-gpgv @{exec_path} {
   /etc/ r,
   /root/ r,
 
-        /var/lib/apt/lists/{,**} r,
+        /{var/lib/ubuntu-advantage/apt-esm/,}var/lib/apt/lists/{,**} r,
```
roddhjav, line 72: I think someting like this makes more sense:
```
/var/lib/apt/lists/{,**} r,
/var/lib/ubuntu-advantage/apt-esm/{,**} r,
```

roddhjav:
Thanks. Merged

Issue #128, opened 2023-02-27
/{usr/,}bin/keepassxc-proxy  rPx,  doesn't work in browsers


curiosityseeker:
In Firefox the

`/{usr/,}bin/keepassxc-proxy     rPx,`

rule and in Brave the 

`/{usr/,}bin/keepassxc-proxy   rPUx,`

rule from the `chromium` abstraction result in the following errors:

`firefox exec /usr/bin/keepassxc-proxy info="no new privs" comm="DOM Worker" requested_mask=x denied_mask=x error=-1`

and

`brave exec /usr/bin/keepassxc-proxy info="no new privs" comm=ThreadPoolForeg requested_mask=x denied_mask=x error=-1`

although the respective profile exists.

Only 

`/{usr/,}bin/keepassxc-proxy   rix,`

works flawlessly. Is there another way to overcome this problem?

roddhjav:
There is sadly no better approach (with the current version of apparmor). Apparmor sees the `keepassxc-proxy` profile with more privileges than the `chromium` profile and does not allow to move to it. The only solution is to give more privileges to the calling profile.

curiosityseeker:
Thanks, that confirms my view. So this should be changed. I'm going to make a PR.

PR #129, opened 2023-03-02
Solve issue #128

curiosityseeker:


nobody43:
I think it should be explicitly commented as a temporary solution.

roddhjav:
Thanks merged !

PR #130, opened 2023-03-03
Adding capability net_admin

curiosityseeker:
capability net_admin is needed to prevent breaking the boot process on Arch Linux with KDE.


roddhjav:
Thanks. Merged

PR #140, opened 2023-04-03
Additonal rules for various profiles

curiosityseeker:


roddhjav:
Nice work, thanks a lot. Merged!

PR #141, opened 2023-04-04
Additional rules for various profles

curiosityseeker:


apparmor.d/groups/freedesktop/pipewire
```diff
@@ -58,6 +58,9 @@ profile pipewire @{exec_path} flags=(attach_disconnected) {
 
   / r,
   /.flatpak-info r,
+  
+  owner @{user_config_dirs}/pipewire/pipewire-pulse.conf r,
+  owner @{user_config_dirs}/pipewire/pipewire.conf r
```
roddhjav, line 63: I did not see before merging. Can you fix this: `r,`

PR #145, opened 2023-04-06
Update run-parts

curiosityseeker:


apparmor.d/profiles-m-r/run-parts
```diff
@@ -15,10 +15,19 @@ profile run-parts @{exec_path} {
...
+  /{usr/,}bin/snapper rix,
+  
+    
   /usr/share/update-notifier/notify-reboot-required   rPx,
   /usr/share/update-notifier/notify-updates-outdated  rPx,
 
   /etc/ r,
+  /etc/conf.d/snapper{,**}                             r,
+  /etc/nsswitch.conf                                   r,
+  /etc/passwd                                          r,
```
roddhjav, line 29: Use the nameservice-strict abstraction instead.

roddhjav:
The abstraction file should go in the profile with the other abstraction: `include <abstractions/nameservice-strict>`

Also, for future PR, please do not edit the file directly on Github:
- Write your change locally
- Test the change by building the package, therefore you will catch all compilation issue
- Then commit and push to GitHub
-

curiosityseeker:
I understand and am sorry for the inconveniences. So right now I'm not going to make any further commits. I will rather try to make myself more familiar with github and find a proper workflow for me. I hope that I can be back some time in the future.

curiosityseeker:
Is still anything missing?

PR #161, opened 2023-05-08
Brave: adjust @{exec_path}

curiosityseeker:
The path in Ubuntu is:
/opt/brave.com/brave/brave

The path in Arch is:
/opt/brave-bin/brave

That's why Brave was not confined on Arch.

### DIFF
--- a/apparmor.d/groups/browsers/brave
+++ b/apparmor.d/groups/browsers/brave
@@ -13,7 +13,7 @@ include <tunables/global>
 @{chromium_config_dirs} = @{user_config_dirs}/BraveSoftware/Brave-Browser{,-Beta,-Dev}
 @{chromium_cache_dirs} = @{user_cache_dirs}/BraveSoftware/Brave-Browser{,-Beta,-Dev}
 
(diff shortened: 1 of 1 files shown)


roddhjav:
Haaa... thanks.

In this case, instead of changing `exec_path`, you should change `chromium_lib_dirs` to  `/opt/brave.com/@{chromium_name} /opt/brave-bin/`

curiosityseeker:
Yes, this looks like a better solution. However, after doing what you suggested and reloading AppArmor I got an error:

```
Job for apparmor.service failed.
See "systemctl status apparmor.service" and "journalctl -xeu apparmor.service" for details.
```
And systemctl status apparmor reports:
```
apparmor.service: Control process exited, code=exited, status=1/FAILURE
Reload failed for Load AppArmor profiles.
profile has merged rule with conflicting x modifiers
FEHLER Verarbeitung der Regexs f?r Profil >>brave<<, Laden gescheitert
Error: At least one profile failed to load
```
So far I haven't found the cause.

curiosityseeker:
Any idea how to solve this?

PR #164, opened 2023-05-24
Update mount, kde-powerdevil and mullvad-daemon

curiosityseeker:


apparmor.d/profiles-m-r/mount
```diff
@@ -9,11 +9,12 @@ abi <abi/3.0>,
 include <tunables/global>
 
 @{exec_path} = /{usr/,}{s,}bin/mount
-profile mount @{exec_path} {
+profile mount @{exec_path} flags=(attach_disconnected) {
   include <abstractions/base>
   include <abstractions/consoles>
   include <abstractions/disks-write>
   include <abstractions/nameservice-strict>
+  include <abstractions/user-tmp>
```
roddhjav, line 17: What user-tmp are required. This abstraction is usually not needed.

curiosityseeker: After the Arch Linux git migration I had landed in systemd's emergency mode for whatever reason. I was able to fix that - but thereafter I saw requests from `mount` to files in /tmp. Unfortunately I can't remember which ones. 

Since then I cannot reproduce these requests, so I'm going to revert this change.

PR #177, opened 2023-07-08
Various updates

curiosityseeker:


apparmor.d/groups/browsers/firefox
```diff
@@ -134,7 +134,7 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
   @{firefox_lib_dirs}/{,**}             r,
   @{firefox_lib_dirs}/*.so              mr,
   @{firefox_lib_dirs}/crashreporter     rPx,
-  @{firefox_lib_dirs}/glxtest          rPUx,
```
roddhjav, line 137: Remove this, this is rPUx on purpose, the profile is a wip.

apparmor.d/groups/browsers/firefox
```diff
@@ -134,7 +134,7 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
...
+  @{firefox_lib_dirs}/glxtest           rix,
```
roddhjav, line 137: Remove this, this is rPUx on purpose, the profile is a wip.

apparmor.d/profiles-m-r/needrestart
```diff
@@ -24,12 +24,16 @@ profile needrestart @{exec_path} flags=(attach_disconnected) {
   @{exec_path} mrix,
 
   /{usr/,}bin/{,ba,da}sh                        rix,
+  /{usr/,}bin/cat                               rix,
   /{usr/,}bin/dpkg-query                        rpx,
   /{usr/,}bin/fail2ban-server                   rPx,
+  /{usr/,}bin/gettext                           rix,
   /{usr/,}bin/locale                            rix,
+  /{usr/,}bin/pacman                            rix,
```
roddhjav, line 32: That cannot be in `rix`.

apparmor.d/groups/browsers/firefox
curiosityseeker: Hm, I'm not sure that I understand. rPUx is causing an error with aa-log (and aa-logprof) every time. How is this related with the profile being a wip? Shouldn't that rule be adjusted, at least, temporarily until a final solution is found?

roddhjav: `rPUx` means it allow to transition to a profile if it exist or unconfined if it does not exist. I do this while I am working on the new `firefox-glxtest` profile. This allow me to transition to the wip profile while other can use the profile without break.

PR #183, opened 2023-07-26
Adding /dev/tty[0-9]* and /dev/pts/[0-9]* to various profiles; update kded5 and reflector

curiosityseeker:


roddhjav:
Can you use the `consoles` abstraction instead of `/dev/pts/[0-9]*` and `/dev/tty`

curiosityseeker:
I was seriously considering this. However, aa-log (and aa-logprof) suggested the `owner` qualifier for `/dev/pts/[0-9]*` in every case, and that abstraction does not offer this. If you think that's okay nevertheless, I can change it as you said.

roddhjav:
Good point. `tty` mediation is a mess anyway. Keeping the `owner` will be better. Merged. Thanks.

Issue #131, opened 2023-03-14
File conflicts with upstream AppArmor: abstractions/trash and tunables/xdg-user-dirs


cboltz:
`abstractions/trash` and `tunables/xdg-user-dirs` also exist in the upstream default profile set, but with completely different rules.

Please find a way to avoid this file conflict, for example by moving these two files to a `*.d/` directoy - or by upstreaming your additions ;-)

roddhjav:
Upstreaming all the abstractions is is planned. At some point, we could even consider upstreaming most of the profiles. But for now, the project is still in early stage where I test (and break) things.

Currently, the packages [displace the files](https://github.com/roddhjav/apparmor.d/blob/main/debian/apparmor.d.displace) to prevent conflict issues.

Regarding `abstractions/trash`, this is still a work in progress (see #94), and right now it is not ready at all for upstream. I ended up overwriting the trash abstraction, as some rules were conflicting.

Regarding the `xdg-user-dirs`, I overwrited the existing one for historic purposes that may not be valid anymore. Actually I would like to see your opinion regarding the new variables defined in `tunables/xdg-user-dirs`. Do you think some could fit upstream? I mean, some are really usefull, some are not. For example `user_tmp_dirs` is useless (and probably not a good idea anyway).

cboltz:
You might be interested in https://build.opensuse.org/package/show/home:cboltz/apparmor.d ;-)

I had to add some workarounds to avoid conflicts - see the spec file and the patch I added, which should answer your questions about the conflicting files. (Maybe you can apply some of my workarounds to your repo?)

In case you wonder: Moving `tunables/extend` to `tunables/etc.d/` is sort of a workaround to load it at the right time, until we have an upstream solution to extend `tunables/global`.

So far, your profiles seem to work. However, I installed the apparmor.d package only some minutes ago. I'll tell you in some days if I see things that need additions to your profiles.

Regarding the new variables - if you have a good use case for them, there are good chances to get them upstream. (No promises ;-)

roddhjav:
Nice, thanks a lot. I will move `tunables/extend` to `tunables/etc.d/`. And try removing all the conflicts.

Enjoy the profile ;). Please fell free to report any issue with them. They are still pretty much a work in progress.

roddhjav:
[...]

PR #186, opened 2023-07-26
Fix xdg user dirs

ShellCode33:
This PR aims to allow users to reuse existing variables in their custom config. I also believe that `xdg-user-dirs.d` is more suited than `home.d` for such a config file.

Currently in my `/etc/apparmor.d/tunables/xdg-user-dirs.d/local` I'm unable to do:

```
@{user_config_dirs}+=@{HOME}/.dotfiles/@{XDG_CONFIG_HOME}
```

Because `XDG_CONFIG_HOME` does not exist in that scope.

As stated in #131, the idea would be to create a subfolder `/etc/apparmor.d/tunables/xdg-user-dirs.d/apparmor.d.extend` to prevent upstream from sourcing the user config and source your `apparmor.d` first. Then all your `apparmor.d` has to do is to source the user folder:

```
include if exists <tunnables/xdg-user-dirs.d/apparmor.d.extend>
```

I also renamed all the `*_HOME` to `*_DIR` for consistence, upstream seems to use that convention:

```
# ------------------------------------------------------------------
#
#    Copyright (C) 2014 Canonical Ltd.
#
#    This program is free software; you can redistribute it and/or
#    modify it under the terms of version 2 of the GNU General Public
#    License published by the Free Software Foundation.
#
# ------------------------------------------------------------------

# Define the common set of XDG user directories (usually defined in
# /etc/xdg/user-dirs.defaults)
@{XDG_DESKTOP_DIR}="Desktop"
@{XDG_DOWNLOAD_DIR}="Downloads"
@{XDG_TEMPLATES_DIR}="Templates"
@{XDG_PUBLICSHARE_DIR}="Public"
@{XDG_DOCUMENTS_DIR}="Documents"
@{XDG_MUSIC_DIR}="Music"
@{XDG_PICTURES_DIR}="Pictures"
@{XDG_VIDEOS_DIR}="Videos"

# Also, include files in tunables/xdg-user-dirs.d for site-specific adjustments
# to the various XDG directories
include <tunables/xdg-user-dirs.d>
```

Also, I was wondering about all the :

```
@{user_documents_dirs}=@{HOME}/@{XDG_DOCUMENTS_DIR} @{MOUNTS}/@{XDG_DOCUMENTS_DIR}
@{user_download_dirs}=@{HOME}/@{XDG_DOWNLOAD_DIR} @{MOUNTS}/@{XDG_DOWNLOAD_DIR}
@{user_music_dirs}=@{HOME}/@{XDG_MUSIC_DIR} @{MOUNTS}/@{XDG_MUSIC_DIR}

[...]
```

Don't you think it would be easier to just append to the existing `XDG_*` ? The above could become:

```
@{XDG_DOCUMENTS_DIR}+=@{MOUNTS}/@{XDG_DOCUMENTS_DIR}
@{XDG_DOWNLOAD_DIR}+=@{MOUNTS}/@{XDG_DOWNLOAD_DIR}
@{XDG_MUSIC_DIR}+=@{MOUNTS}/@{XDG_MUSIC_DIR}

[...]
```

And the user could also add its own:

```
@{XDG_CONFIG_DIR}+=.dotfiles/@{XDG_CONFIG_DIR}
@{XDG_DATA_DIR}+=.dotfiles/@{XDG_DATA_DIR}

[...]
```

I think it would be much easier and it would not require creating new variables.

In the profiles you would then have to use `@{HOME}` everywhere :

```
@{HOME}/@{XDG_DOCUMENTS_DIR}/salut rw,
```

And as a bonus, it would allow you to use some upstream abstractions that already depend on `XDG_*` variables such as [this one](https://gitlab.com/apparmor/apparmor/-/blob/17a521ff5051d4e36c38d2cb760d0d97773f5a8d/profiles/apparmor.d/abstractions/user-write).

### DIFF
--- a/apparmor.d/groups/freedesktop/xdg-desktop-portal-gtk
+++ b/apparmor.d/groups/freedesktop/xdg-desktop-portal-gtk
@@ -162,7 +162,7 @@ profile xdg-desktop-portal-gtk @{exec_path} {
   owner @{HOME}/ r,
   owner @{HOME}/.* r,
   owner @{HOME}/.icons/{,**} r,
-  owner @{HOME}/@{XDG_DATA_HOME}/ r,
+  owner @{HOME}/@{XDG_DATA_DIR}/ r,
 
   owner /tmp/runtime-*/xauth_?????? r,
 
--- a/apparmor.d/groups/freedesktop/xdg-permission-store
+++ b/apparmor.d/groups/freedesktop/xdg-permission-store
@@ -46,7 +46,7 @@ profile xdg-permission-store @{exec_path} flags=(attach_disconnected) {
 
   @{exec_path} mr,
 
-  @{HOME}/@{XDG_DATA_HOME}/flatpak/db/gnome rw,
+  @{HOME}/@{XDG_DATA_DIR}/flatpak/db/gnome rw,
 
   owner @{user_share_dirs}/flatpak/ w,
   owner @{user_share_dirs}/flatpak/db/ rw,
--- a/apparmor.d/profiles-m-r/man
+++ b/apparmor.d/profiles-m-r/man
@@ -104,10 +104,10 @@ profile man_filter {
   # there's no harm in allowing wide read access here since the worst it can
   # do is feed data to the invoking man process.
         /usr/** r,
-  owner @{HOME}/@{XDG_DATA_HOME}/** r,
+  owner @{HOME}/@{XDG_DATA_DIR}/** r,
   owner @{user_projects_dirs}/** r,
   owner @{user_cache_dirs}/** r,
-  owner @{MOUNTS}/*/@{XDG_DATA_HOME}/** r,
+  owner @{MOUNTS}/*/@{XDG_DATA_DIR}/** r,
 
   /var/cache/man/** w,
 
--- a/apparmor.d/profiles-s-z/syncthing
+++ b/apparmor.d/profiles-s-z/syncthing
@@ -30,7 +30,7 @@ profile syncthing @{exec_path} {
   /usr/share/mime/globs2 r,
 
   owner @{HOME}/ r,
-  owner @{HOME}/@{XDG_DATA_HOME}/syncthing/{,**} rwk,
+  owner @{HOME}/@{XDG_DATA_DIR}/syncthing/{,**} rwk,
   owner @{user_config_dirs}/syncthing/{,**} rwk,
 
   /home/ r,
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -31,19 +31,19 @@
 @{XDG_PASSWORD_STORE_DIR}=".password-store"
 
 # Definition of local user configuration directories
-@{XDG_CACHE_HOME}=".cache"
-@{XDG_CONFIG_HOME}=".config"
-@{XDG_DATA_HOME}=".local/share"
-@{XDG_STATE_HOME}=".local/state"
-@{XDG_BIN_HOME}=".local/bin"
-@{XDG_LIB_HOME}=".local/lib"
+@{XDG_CACHE_DIR}=".cache"
+@{XDG_CONFIG_DIR}=".config"
+@{XDG_DATA_DIR}=".local/share"
+@{XDG_STATE_DIR}=".local/state"
+@{XDG_BIN_DIR}=".local/bin"
+@{XDG_LIB_DIR}=".local/lib"
 
 # Full path of the user configuration directories
-@{user_cache_dirs}=@{HOME}/@{XDG_CACHE_HOME}
-@{user_config_dirs}=@{HOME}/@{XDG_CONFIG_HOME}
-@{user_state_dirs}=@{HOME}/@{XDG_STATE_HOME}
-@{user_bin_dirs}=@{HOME}/@{XDG_BIN_HOME}
-@{user_lib_dirs}=@{HOME}/@{XDG_LIB_HOME}
+@{user_cache_dirs}=@{HOME}/@{XDG_CACHE_DIR}
+@{user_config_dirs}=@{HOME}/@{XDG_CONFIG_DIR}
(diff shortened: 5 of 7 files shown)


roddhjav:
Hi, thank a lot for this PR (and for for the sponsorship!!)

> I also believe that xdg-user-dirs.d is more suited than home.d for such a config file.

`xdg-user-dirs.d` is better, but this is a hack to ensure the variable load before other in the directory (#131)

> Currently in my /etc/apparmor.d/tunables/xdg-user-dirs.d/local I'm unable to do:
> ```
> @{user_config_dirs}+=@{HOME}/.dotfiles/@{XDG_CONFIG_HOME}
> ```
> Because `XDG_CONFIG_HOME` does not exist in that scope.
> 

I do not get this part,  `XDG_CONFIG_HOME` exist, this is the whole point of defining it in `home.d` as it is loaded before `xdg-user-dirs.d`.

> I also renamed all the *_HOME to *_DIR for consistence, upstream seems to use that convention:

`XDG_*_HOME` variable are named after [XDG_Base_Directory](https://wiki.archlinux.org/title/XDG_Base_Directory). However, XDG Base Directory are full path, while `*_DIR` are just directory name. So it make sense to rename them. :+1: 

> Also, I was wondering about all the :
> ```
> @{user_documents_dirs}=@{HOME}/@{XDG_DOCUMENTS_DIR} @{MOUNTS}/@{XDG_DOCUMENTS_DIR}
> ```
> ...

The logic is that `@{user_*_dirs}` are full path (`/home/user/Documents /mnt/*/Documents`), while `@{XDG_*_DIR}` are just directory name (`Documents`) this allow to support directory on mount-point  as well as classic homedir. Your proposal would break.

Upstreaming is another subject for later (and the variables may be uptreamed first anyway)

ShellCode33:
Ok so I tested again from scratch by doing a clean install, and I think I got confused at some point. The error I had was probably due to me using `XDG_CONFIG_DIR` instead of `XDG_CONFIG_HOME`, I guess journalctl was telling me that `XDG_CONFIG_DIR` was not defined and for some reasons I thought it was due to the fact that variables coming from `home.d/apparmor.d` were not propagated to `xdg-user-dirs.d/`. I should probably have used `@{user_config_dirs}` anyway... But it works fine indeed using `XDG_CONFIG_HOME`, sorry for that.

However I still think that using the `tunnables/xdg-user-dirs.d/something/` subdirectory hack to prevent upstream breakage would make more sense than using `home.d/` but I also totally get your point in #131 about breaking everyone's setup (even though there's probably something to do with symlinks for a smooth transition). But it's fine like this I guess, this PR was mainly about addressing the `XDG_CONFIG_DIR` part I messed up.

So what do you want to do next ? I can force push my branch to only keep the commits related to the renaming of `_HOME` to `_DIR`. But I believe it will also break some people's setup. I feel like any modification will break someone's setup anyway...

We could do something like this to ensure retro-compatibility:

```
@{XDG_CACHE_DIR}=".cache"
@{XDG_CONFIG_DIR}=".config"
@{XDG_CACHE_HOME}=@{XDG_CACHE_DIR}
@{XDG_CONFIG_HOME}=@{XDG_CONFIG_DIR}
```

But the more you want to keep changes retro-compatible, the more it will clutter the config files, it's your choice to make !

ShellCode33:
What if I want to override variables defined in `/etc/apparmor.d/tunables/xdg-user-dirs.d/apparmor.d` though ?

I might want to have multiple documents folders and put something like this in my `/etc/apparmor.d/tunables/xdg-user-dirs.d/local` :

```
@{user_documents_dirs}+=/somewhere/
```

roddhjav:
You can do the rename, that would not break too much thing anyway. The problem solved by #131 is that apparmor load files in folder in a random way. It is is what could have messed up a lot of thing if we were going to move definition. This is also why we have to keep some variable definition in `home.d` as they must be loaded early. And this is exactly the same issue that is the cause of your error message (sometime `tunables/xdg-user-dirs.d/apparmor.d` will be loaded first and apparmor will compile the profile, sometime it will not, and apparmor will fail). So here, yes the solution is to add an include in `tunables/xdg-user-dirs.d/apparmor.d`. 

So to sum up, do a force push with:
1. `_HOME` to `_DIR`
2. Add a `include if exists <tunnables/xdg-user-dirs.d/apparmor.d.d>` in `tunnables/xdg-user-dirs.d/apparmor.d`

ShellCode33:
Done !

May I ask why some `@{user_*}` variables are in `tunables/xdg-user-dirs.d/apparmor.d` while others are in `tunables/home.d/apparmor.d` ?

ShellCode33:
How about, instead of introducing new variables like `XDG_CONFIG_DIR`, `XDG_DATA_DIR`, etc. which are not defined by the Freedesktop standard we keep the `XDG_*_HOME` you aleady have, make them full path (as dictated by the Freedesktop standard), and change all the profiles to use `@{XDG_*_HOME}` instead of `@{HOME}/@{XDG_*_DIR}` ? I think it would make more sense. I can reflect that in this PR if you want. But then `@{XDG_*_HOME}` would be a duplicate of your `@{user_*_dirs}` variables. 

So I guess there are two ways of doing that:

- We keep your existing `${user_*_dirs}` but we completely remove occurrences of @{XDG_*_HOME}
- Or we completely remove the `${user_*_dirs}` (only the ones that have a Freedesktop equivalent) and we use @{XDG_*_HOME} in the profiles instead, but remove the `@{HOME}` prefix

roddhjav:
I merged the PR as it is right now. Further variable improvement can be done in a new PR.

Also, I think both `${user_*_dirs}` and `@{XDG_*_HOME}` can have a use. In Apparmor, XDG are relative dirname, while `user_` are full path  (including mounts and user addition)

ShellCode33:
Alright, thanks for taking the time to review and merge it !

I'm really hoping that https://gitlab.com/apparmor/apparmor/-/merge_requests/1081 will be merged, I'm sure it will help you as well

PR #189, opened 2023-08-05
New tunables

nobody43:
I haven't noticed any measurable performance impact on profile's loading.
Thrown in a couple of unrelated edits.
Thoughts?


apparmor.d/groups/bus/ibus-dconf
```diff
@@ -33,16 +33,16 @@ profile ibus-dconf @{exec_path} flags=(attach_disconnected) {
   /etc/dconf/db/ibus r,
   /etc/dconf/profile/ibus r,
 
-  /var/lib/gdm{3,}/.config/ibus/bus/{,@{hex}-unix-wayland-[0-9]*} r,
```
nobody43, line 36: Can we avoid such constructs? It's cryptic even for a professional, let alone novice.

roddhjav:
I am 100% in with this change, just allow a bit of time to merge it.

@cboltz do you have any comment on these new variables?

Some note:
- I would rather have `@{number}` named `@{int}`
- While you are at it, can you also remove the non-working time & date variable too

apparmor.d/groups/gnome/gnome-software
```diff
@@ -86,7 +86,7 @@ profile gnome-software @{exec_path} {
 
   owner /tmp/ostree-gpg-*/ rw,
   owner /tmp/ostree-gpg-*/** rwkl -> /tmp/ostree-gpg-*/**,
-  owner /tmp/#[0-9]* rw,
```
nobody43, line 89: Assumptions

nobody43:
Now, where do we apply `@{int}`?
a) On every path
b) On user-writable paths

nobody43:
A couple of dozen careful assumptions. Might break something but I think it won't.

apparmor.d/profiles-s-z/xauth
```diff
@@ -31,11 +31,11 @@ profile xauth @{exec_path} {
   owner /tmp/serverauth.*-n rw,
   owner /tmp/serverauth.*   rwl -> /tmp/serverauth.*-n,
 
-  owner /tmp/runtime-*/xauth_?????? r,
+  owner /tmp/runtime-*/xauth_@{rand6} r,
 
-  owner @{run}/user/@{uid}/xauth_?????? rw,
-  owner @{run}/user/@{uid}/xauth_??????-c w,
-  owner @{run}/user/@{uid}/xauth_??????-l wl,
```
nobody43, line 38: On a second thought, these might cause problems, because I expanded ambiguous `/xauth_*` to unequivocal `/xauth_@{rand6}` in other profiles.

apparmor.d/tunables/multiarch.d/apparmor.d
```diff
@@ -6,15 +6,33 @@
...
 
 # Hexadecimal
-@{hex}=[0-9a-fA-F]*
+@{hex}=@{h}*@{h}
 
 # Date and time
-@{date}=[0-9][0-9][0-9][0-9]-[1-12]-[1-31]
-@{time}=[1-24]-[0-60]-[0-60]
+@{date}=[0-9][0-9][0-9][0-9]-[1-9]{,[0-2]}-[1-9]{,[0-9]}
+@{time}=[1-9]{,[0-9]}-[0-9]{,[0-9]}-[0-9]{,[0-9]}
```
cboltz, line 35: These two variables look wrong (and if their previous version is also wrong). IMHO they should be

    @{date}=[0-2][0-9][0-9][0-9]-[01][0-9]-[0-3][0-9]
    @{time}={[0-2],}[0-9]-[0-5][0-9]-[0-6][0-9]

I assume the goal is to support iso date format (`2023-08-08`) and time in `5-42-42` format (with optional leading zero for the hours.

My @{date} will only support years up to 2999 - feel free to adjust the first digit range to make it more future-proof ;-)

In case you wonder why I picked `[0-6][0-9]` for the seconds - search for "leap second".

nobody43:
Thanks for the suggestions!

apparmor.d/groups/bus/dbus-daemon
```diff
@@ -77,8 +77,6 @@ profile dbus-daemon @{exec_path} flags=(attach_disconnected) {
   owner @{user_share_dirs}/dbus-1/{,**} r,
         @{user_share_dirs}/icc/{,edid-*} r,
 
-  owner /tmp/dbus-[0-9a-zA-Z]* rw,
```
nobody43, line 80: Abstraction.

roddhjav:
Merged with 557d905543e89a458fa6caab0fdd5c288095c923

PR #192, opened 2023-08-08
Child open: add some executables and complete kde-open5

curiosityseeker:


apparmor.d/groups/children/child-open
```diff
@@ -86,6 +90,11 @@ profile child-open {
   @{bin}/viewnior         rPUx,
   @{bin}/vlc               rPx,
   @{bin}/xarchiver         rPx,
+  @{lib}/libreoffice/program/{soffice,soffice.bin,oosplash} rPUx,
+
+  owner @{user_config_dirs}/menus/{,**} r,
+
+  owner @{run}/user/@{uid}/kioclient*.[0-9]*.kioworker.socket rwl -> @{run}/user/@{uid}/#[0-9]*[0-9],
```
roddhjav, line 98: I think this should be better placed in `kde-open5.d/complete`

apparmor.d/groups/children/child-open
```diff
@@ -86,6 +90,11 @@ profile child-open {
...
```
roddhjav, line 97: I think this should be better placed in `kde-open5.d/complete`

PR #205, opened 2023-09-03
New profile: mariadbd_akonadi

curiosityseeker:
A similar approach is available on https://invent.kde.org/pim/akonadi/-/tree/master/apparmor?ref_type=heads where also @cboltz contributed.

This profile has some additional rules and is formatted according to the guidelines of this project.

### DIFF
--- a/apparmor.d/groups/kde/mariadbd_akonadi
+++ b/apparmor.d/groups/kde/mariadbd_akonadi
@@ -5,13 +5,16 @@ abi <abi/3.0>,
 @{exec_path} = @{bin}/mariadb{,-*} @{bin}/mariadbd @{bin}/mysql @{bin}/mysqld{,*}
 profile mariadbd_akonadi @{exec_path} flags=(complain) {
   include <abstractions/base>
-  include <abstractions/bash>
   include <abstractions/consoles>
-  include <abstractions/nameservice>
+  include <abstractions/nameservice-strict>
   include <abstractions/openssl>
   
   capability setgid,
   capability setuid,
+
+  network inet dgram,
+  network inet6 dgram,
+  network netlink raw,
 
   signal (receive) set=(kill, term) peer=akonadiserver,
     
@@ -43,7 +46,7 @@ profile mariadbd_akonadi @{exec_path} flags=(complain) {
   
   @{sys}/block/ r,
   @{sys}/devices/system/cpu/ r,
-  @{sys}/devices/pci[0-9]*/**/{dev,rotational} r,
+  @{sys}/devices/@{pci}/{dev,rotational} r,
   @{sys}/devices/virtual/block/zram@{int}/queue/rotational r,
   @{sys}/devices/virtual/block/zram@{int}/dev r,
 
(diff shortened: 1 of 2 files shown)


apparmor.d/groups/kde/mariadbd_akonadi
```diff
@@ -0,0 +1,54 @@
+abi <abi/3.0>,
+
+#include <tunables/global>
+
+@{exec_path} = @{bin}/mariadb{,-*} @{bin}/mariadbd @{bin}/mysql @{bin}/mysqld{,*}
+profile mariadbd_akonadi @{exec_path} flags=(complain) {
+  include <abstractions/base>
+  include <abstractions/bash>
+  include <abstractions/consoles>
+  include <abstractions/nameservice>
```
roddhjav, line 10: Use nameservice-strict. And add network access in the profile.

apparmor.d/groups/kde/mariadbd_akonadi
```diff
@@ -0,0 +1,54 @@
+abi <abi/3.0>,
+
+#include <tunables/global>
+
+@{exec_path} = @{bin}/mariadb{,-*} @{bin}/mariadbd @{bin}/mysql @{bin}/mysqld{,*}
+profile mariadbd_akonadi @{exec_path} flags=(complain) {
+  include <abstractions/base>
+  include <abstractions/bash>
```
roddhjav, line 8: The bash abstraction is only needed for interactive shell. So not here.

roddhjav:
What don't you simply use the profiles they provide?

apparmor.d/groups/kde/mariadbd_akonadi
```diff
@@ -0,0 +1,54 @@
+abi <abi/3.0>,
+
+#include <tunables/global>
+
+@{exec_path} = @{bin}/mariadb{,-*} @{bin}/mariadbd @{bin}/mysql @{bin}/mysqld{,*}
```
roddhjav, line 5: That won't work. Doing so you are going to confine a mariadbd server. Not the database of akonadi.

apparmor.d/groups/kde/mariadbd_akonadi
curiosityseeker: That abstraction was in the original profile. I have to check the consequences once removed.

apparmor.d/groups/kde/mariadbd_akonadi
curiosityseeker: I'm not sure that I understand. In my understanding AppArmor profiles do not confine data(bases) (unless specific data is protected by including,e.g., the private-files or sensitive-home abstractions or adding specific deny rules) but rather applications are confined by allowing them to do/access only what's necessary. So how is this profile (and the original one) different from all the other ones? 

FWIW, I've been using this profile for quite some time in enforce mode:
`/usr/bin/mariadbd (8323) mariadbd_akonadi`

I'll remove the bash abs and replace the nameservice abs with nameservice-strict  and see what's necessary to add.

apparmor.d/groups/kde/mariadbd_akonadi
curiosityseeker: That was in the original profile.

curiosityseeker:
Is there still something missing? If you don't want to accept this PR I can delete it, no problem.

roddhjav:
Sorry for the delay. I would rather focus on profile that need to be done than on profile that are already packaged (plus they would be some change to be done to actually merge it).

PR #209, opened 2023-09-05
Various updates

curiosityseeker:


apparmor.d/profiles-a-f/blkid
```diff
@@ -35,5 +35,8 @@ profile blkid @{exec_path} {
   # For the EVALUATE=scan method
   @{PROC}/partitions r,
 
+  /dev/pts/@{int} rw,
```
roddhjav, line 38: Already in `consoles`?

dists/flags/main.flags
```diff
@@ -32,6 +32,8 @@ avahi-publish complain
 avahi-resolve complain
 avahi-set-host-name complain
 baloo complain
+blkid attach_disconnected
+btrfs attach_disconnected
```
roddhjav, line 36: it should be: attach_disconnected,complain

roddhjav: Or don't put them here at all. The purpose of this file is to overwrite flag by new (temporary) ones. So, as these profiles are usually enforced. There is no need to add them here.

curiosityseeker: > Or don't put them here at all. The purpose of this file is to overwrite flag by new (temporary) ones. So, as these profiles are usually enforced. There is no need to add them here.

Sorry, I don't understand. The attach_disconnected flag has to be added to main.flags in any case?

curiosityseeker: Ah, I see.

curiosityseeker: Just to clarify: this means that **only** for profiles which are still in `complain` mode additional flags have to be added to `main.flags` and not for profiles in `enforce` mode?

roddhjav: Yes, only for profile that should stay in complain mode no matter how the project is compiled.

roddhjav:
Thanks, merged!

PR #212, opened 2023-09-09
Fix access to /tmp/user/@{uid} for Firefox and new profile for MPD

yukiteruamano:
Firefox require access to /tmp/user/@{uid}/ for downloads in Firefox ESR for actual Debian Stable (FirefoxESR-102.15.0esr-1-deb12u1)

Tested and deployed in my actual workstation. Thanks for this work!


roddhjav:
Thanks a lot for this PR. Merged!

PR #214, opened 2023-09-11
Update gitignore and better defaults for mpv

yukiteruamano:
Add access to GPU ids (amdgpu) for hwaccel in mpv, and better defaults for user_config in this profile.

Update gitignore and debian/control


roddhjav:
Looks good. Thanks a lot. Merged

PR #215, opened 2023-09-11
Fix access to /tmp folders for Firefox ESR in Debian

yukiteruamano:
Firefox ESR in Debian Stable require access to additional folders in /tmp when use a configuration with multiple profiles.


roddhjav:
Thanks, merged

PR #231, opened 2023-10-20
Update pipewire

curiosityseeker:
Necessary after the recent pipewire update, otherwise audio devices are no longer available.


roddhjav:
Thanks. merged

PR #235, opened 2023-10-26
fix gnome slow start up

monsieuremre:
Gentlemen, it is with great pleasure to inform you that I have been able to locate the source of the notorious slow start up problem. This unironically fixes the slow start up on gnome. Almost everything under @{lib} is directories. If we inherit our profile instead of looking for seperate profiles for each and every one of them, it fixes the issue. Maybe this solution can be improved further. But yes, this is the problem, and there is the solution.

### DIFF
--- a/apparmor.d/groups/bus/dbus-daemon
+++ b/apparmor.d/groups/bus/dbus-daemon
@@ -44,7 +44,7 @@ profile dbus-daemon @{exec_path} flags=(attach_disconnected) {
   @{lib}/{,at-spi2{,-core}/}at-spi2-registryd                             rPx,
   @{lib}/@{multiarch}/tumbler-1/tumblerd                                 rPUx,
   @{lib}/@{multiarch}/xfce[0-9]/xfconf/xfconfd                            rPx,
-  @{lib}/*                                                               rPUx,
+  @{lib}/*                                                                rix,
   @{lib}/atril/atrild                                                     rPx,
   @{lib}/dbus-1*/dbus-daemon-launch-helper                                rPx,
   @{lib}/gnome-shell/gnome-shell-calendar-server                          rPx,


monsieuremre:
I am aware of this. The sole issue here is @{lib}/* is just the directories under lib. They are executable because they are directories. But they obviously don't have profiles. I suspect apparmor looks for their profiles among the 1.5k profiles, for each one of them. This command does not look into subdirectories. It is not /** it is just /*. So a better solution would be delete the rule. Then add @{lib}/binary for each binary we need. We don't need the directories anyway, even if we did, this rules just covers executing (opening) the directory and nothing else.

roddhjav:
> The sole issue here is @{lib}/* is just the directories under lib.

`@{lib}/*` does not cover directories, `@{lib}/*/` would cover directories (and only directories).

> So a better solution would be delete the rule. Then add @{lib}/binary for each binary we need.

I am going to do some test on this. Likely one of the profile starting by dbus is causing the issue.

monsieuremre:
Sorry my understanding was that ```@{lib}``` was ```/usr/lib``` but I see it is in fact ```/usr/libexec```. There seems to be a lot of stuff there. So I don't know how to identify the bad profile. But I can tell it is non of the gnome stuff. Because the problem persists with all gnome profiles disabled. This narrows down our candidates by quite a bit. Hope this helps.

roddhjav:
`@{lib}` is actually both: `/{usr/,}lib{,exec,32,64}`. See https://apparmor.pujol.io/variables/#system-variables

I can confirm that setting `rix` solve the issue. Putting this rule in audit mode shows that on default install not a lot of programs are started from `@{lib}/*`:
```
at-spi-bus-launcher
at-spi2-registryd
dconf-service
epiphany-webapp-provider
gnome-shell-calendar-server
goa-daemon
goa-identity-service
xdg-permission-store
```
Therefore, it limit the search by a lot... even if it is mostly dbus related profile again (`at-spi*`)

roddhjav:
Found the solution: as `at-spi-bus-launcher` starts the [accessibility bus](https://github.com/GNOME/at-spi2-core/blob/main/bus/README.md#who-launches-the-accessibility-bus), it needs to be in the same profile than the dbus-daemon itself. Otherwise, it is strugling to access dbus (don't really understand why, as it had access). It goes the same for `at-spi2-registryd` as it is just the dbus service for accessibility.

Therefore, the clean solution is to remove `at-spi-bus-launcher`  and  `at-spi2-registryd`  and to include them into `dbus-daemon`.

roddhjav:
Fixed by [e99f7de](https://github.com/roddhjav/apparmor.d/commit/e99f7de70310dff4c00ac4d309a81e07e34d47c0). Thanks again for the initial tip.

PR #236, opened 2023-10-26
Partial install script

monsieuremre:
Please do not merge just yet. I got rid of the function and make the functionality work in a standalone script. It probably just works but it needs testing. After testing we can merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,12 +57,7 @@ $(P):
 	@for file in ${TUNABLES}; do \
 		install -Dm0644 "${BUILD}/apparmor.d/tunables/$${file}" "${DESTDIR}/etc/apparmor.d/tunables/$${file}"; \
 	done;
-	@echo "Warning: profile dependencies fallback to unconfined."
-	@for file in ${@}; do \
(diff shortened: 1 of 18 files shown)


monsieuremre:
There was a problem with regex file names like ```wl-{copy,paste}``` and ```mount.*```. I fixed it. Seems to work in all cases. It should now work for everything. Please do test before merging.

monsieuremre:
This time I tested the whole thing. Turns out there was still an error, and I fixed it. Now it can be merged.

roddhjav:
Thanks.

Nitpicking a bit, but why not having a function in this script and using this function for the recursive part (instead of calling the script again). It seems to me this is what you proposed initially.

roddhjav:
Can you have test [e047da9](https://github.com/roddhjav/apparmor.d/pull/236/commits/e047da90003ac09480c90ed97f2e028d861af861). I get the same result than without the function. It may no be perfect still.

roddhjav:
Closing, as more integration is required anyway a full implementation would most likely have to be integrated within the prebuild system.

PR #275, opened 2024-01-16
Tighten `firefox`

nobody43:
Versions tested:
- 121.0.1 (official tar on Ubuntu 22.04)
- 102.15.0esr (from stock repository on Debian 12)

Behavior tested:
- Regular browsing
- File downloads
- Extension adding/removal
- Themes adding/removal

### DIFF
--- a/apparmor.d/groups/browsers/firefox
+++ b/apparmor.d/groups/browsers/firefox
@@ -148,6 +148,8 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
   owner @{user_share_dirs}/applications/userapp-Firefox-@{rand6}.desktop{,.@{rand6}} rw,
   owner @{user_share_dirs}/mime/packages/user-extension-{htm,html,xht,xhtml,shtml}.xml rw,
   owner @{user_share_dirs}/mime/packages/user-extension-{htm,html,xht,xhtml,shtml}.xml.* rw,
+  owner @{user_share_dirs}/sounds/__custom/index.theme r,
+  owner @{user_share_dirs}/sounds/__custom/*.ogg r,
 
   owner @{config_dirs}/ rw,
   owner @{config_dirs}/{extensions,systemextensionsdev}/ rw,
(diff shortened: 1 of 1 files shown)


apparmor.d/groups/browsers/firefox
```diff
@@ -231,6 +240,7 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
   owner /dev/tty@{int} rw, # File Inherit
 
   # Silencer
+  deny capability sys_ptrace,
```
nobody43, line 243: Might interfere with `crashreporter`? Anyway, I think telemetry should be treated as malicious.

roddhjav:
What do you mean by "denies capability of self-update"? There is no such cap, and no Linux software can update by itself anyway.

apparmor.d/groups/browsers/firefox
roddhjav: Remove this. You personal opinion on a software feature (even if I do agree with you here) should not interfere in the profile.

apparmor.d/groups/browsers/firefox
```diff
@@ -172,15 +174,22 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
...
+  owner /tmp/Mozillato-be-removed-cachePurge-??????????????? rwk,
+  owner /tmp/Mozilla@{uuid}-cachePurge-??????????????? rwk,
+  owner /tmp/Mozilla\{@{uuid}\}-cachePurge-??????????????? rwk,
+  owner /tmp/Temp-@{uuid}/{**,} rw,
+  owner /tmp/mozilla-temp-@{int} rw,
+  owner /tmp/@{rand8}.txt w,
+  owner /tmp/tmp-???.xpi rw,
+  owner /tmp/.xfsm-ICE-@{rand6} rw,
+  owner /tmp/tmpaddon r,
+  owner /tmp/* w,  # file downloads (to anywhere)
```
roddhjav, line 192: ?

apparmor.d/groups/browsers/firefox
nobody43: Removed. I actually agree on this, as long as there's pre-compile/post-compile option (looking forward in the future).

apparmor.d/groups/browsers/firefox
nobody43: On every file download, FF attempts to write a file in `/tmp` with arbitrary name, regardless of actual saving directory. Read access is not needed.

nobody43:
> What do you mean by "denies capability of self-update"? There is no such cap, and no Linux software can update by itself anyway.

I'm not talking about Linux capabilities, I should have said 'ability', sorry. 

This repository contain profiles which have an ability of self-updates (not necessary from repositories), e.g.:
- Firefox
- Thunderbird
- Tor browser
- Brave?
- spectre-meltdown-checker.sh

PR #276, opened 2024-01-19
Updating polkit, sddm, unix chkpwd, btrfs, needrestart

curiosityseeker:
Today the boot process in Arch was broken for me. I'm not quite sure what exactly caused that but those modifications fixed it. 

I got those rules from aa-log after booting into a snapshot (I'm using btrfs with snapper).


roddhjav:
Thanks a lot, and sorry for the delay. Merged!

PR #277, opened 2024-01-21
Unhide potential errors

nobody43:
This takes care of two problems:
1. User knows whether installation introduced additional errors
2. Profiles are loaded (it's not clear to a new user)

### DIFF
--- a/debian/apparmor.d.postinst
+++ b/debian/apparmor.d.postinst
@@ -5,6 +5,6 @@
 
 #DEBHELPER#
 
-systemctl reload apparmor.service >/dev/null 2>&1 || true
+systemctl restart apparmor.service > /dev/null || true
 
 exit 0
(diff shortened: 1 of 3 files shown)


roddhjav:
This should already be automatic thanks to: https://github.com/roddhjav/apparmor.d/blob/main/debian/apparmor.d.postinst

nobody43:
It could be inactive, that's stock Ubuntu:
`apparmor.service is not active, cannot reload.`

Anyway, dumping `stderr` obscures any issues (e.g. multiple tunable definition). I'm not sure whether dumping `stdout` is needed.

What do you think of it?

nobody43:
To reproduce:
1. define `@{XDG_SSH_DIR}` in tunables on a clean system
2. proceed with installation instructions

Result:
Installation will be finished successfully, `systemctl reload apparmor.service` errors will be obscured, and system will be in a degraded state (not all profiles loaded, regardless of the error).

roddhjav:
Thanks. Merged, you are right, error will be easier to catch this way.

PR #288, opened 2024-02-02
Mutt and Abook profiles

doublez13:
AppArmor profile for mutt.

There are an innumerable number of mutt configurations. This AA profile is intended to work with a stock mutt config. Any customizations should be placed in local/mutt

### DIFF
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -32,7 +32,6 @@ profile mutt @{exec_path} {
   @{bin}/sendmail                rPUx,
   @{lib}/sendmail/sendmail rPUx,
   @{bin}/ispell                  rPUx, 
-  # TODO: Add a profile for abook (Most distros don't ship this anymore though)
   @{bin}/abook                   rPUx,
   @{bin}/mutt_dotlock             rix,
   # Misc mutt scripts
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -33,8 +33,7 @@ profile mutt @{exec_path} {
   # Used when saving attachments. Mutt saves attachments in whatever directory
   # it's launched from, which is most likely @{HOME}.
   owner @{HOME}/*           lw,
-  owner @{HOME}/.mutt*/     w,
-  owner @{HOME}/.mutt*/*    w,
+  owner @{HOME}/.mutt*/{,*} w,
   owner @{HOME}/Downloads/* lw,
   # Used When viewing attachments
   owner /tmp/*              lrw,
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -52,7 +52,9 @@ profile mutt @{exec_path} {
 
   /usr/share/terminfo/** r,
 
-  /etc/mime.types r,
+  # Mutt MIME types search path
+  /etc/mime.types           r,
+  owner @{HOME}/.mime.types r,
 
   # Mutt mailcap search path
   /etc/{mutt/,}mailcap   r,
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -9,10 +9,10 @@ include <tunables/global>
 @{exec_path} = @{bin}/mutt
 profile mutt @{exec_path} {
   include <abstractions/base>
+  include <abstractions/consoles>
   include <abstractions/nameservice-strict>
   include <abstractions/openssl>
   include <abstractions/ssl_certs>
-  include <abstractions/consoles>
   include <abstractions/user-download-strict>
   include <abstractions/user-read>
 
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -150,8 +150,9 @@ profile mutt @{exec_path} {
     include <abstractions/base>
     include <abstractions/consoles>
   
-    /{usr/,}bin/less mrix,
-    /{usr/,}bin/more mrix,
+    /{usr/,}bin/less  mr,
+    /{usr/,}bin/more  mr,
+    /{usr/,}bin/pager mr,
     
     owner @{HOME}/.lesshsQ r,
     owner @{HOME}/.local/state/ r,
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -69,14 +69,6 @@ profile mutt @{exec_path} {
 
   owner @{HOME}/.cache/mutt rwk,
 
-  # Used When viewing attachments
-  owner /tmp/*              lrw,
-
-  # Needed to compose a message
-  owner /{var/,}tmp/.mutt*/  rw,
-  owner /{var/,}tmp/.mutt*/* lrwk,
-  owner /{var/,}tmp/mutt*    lrwk,
-
   # Needed for the edit operation.
   owner @{HOME}/ r,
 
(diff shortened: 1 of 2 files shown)


apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2023 Zane Zakraisek <zz@eng.utah.edu>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = /{usr/,}bin/mutt
```
roddhjav, line 9: Use `@{bin}` instead of `/{usr/,}bin`

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+profile mutt @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/nameservice>
```
roddhjav, line 12: Use `nameservice-strict` instead of `nameservice`, and add the required `network` rule in the profile directly.

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+  include <abstractions/openssl>
+  include <abstractions/consoles>
+
+  @{exec_path} mr,
+
+  # Mutt config files
+  /usr/{local/,}share/mutt/Muttrc r,
+  /etc/{mutt/,}Muttrc             r,
+  owner @{HOME}/.mutt/**          r,
```
roddhjav, line 21: Simplify as owner `@{HOME}/.mutt/{,**} rw,`, and remove the later rules (`owner @{HOME}/.mutt*/`).

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+
+  #Needed to open a mailbox (at least an imap one)
+  owner /tmp/.mutt*/  rw,
+  owner /tmp/.mutt*/* lrwk,
+
+  # Might be able to get away without this
+  owner /tmp/mutt* lrwk,
+
+  # Needed for the edit operation.
+  @{HOME} r,
```
roddhjav, line 45:
```suggestion
  @{HOME}/ r,
```

It is a directory, therefore it should end by a `/`.

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+
+  # User mbox
+  owner /var/{spool/,}mail/* rwk,
+  owner @{HOME}/postponed    rwk,
+  owner @{HOME}/sent         rwk,
+  # User maildir
+  owner @{HOME}/{M,m}ail/    rw,
+  owner @{HOME}/{M,m}ail/**  rwk,
```
roddhjav, line 53: Use this instead:
```
  owner @{user_mail_dirs}/ rw,
  owner @{user_mail_dirs}/** rwlk -> @{user_mail_dirs}/**,
```

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+profile mutt @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/openssl>
+  include <abstractions/consoles>
+
+  @{exec_path} mr,
+
+  # Mutt config files
+  /usr/{local/,}share/mutt/Muttrc r,
```
roddhjav, line 19: This should be simplied as:
```
/usr/share/mutt/** r,
```

Therefore other instances of `/usr/share/mutt/` can be removed. Also local files are usually out of scope of apparmor profile (that's not a project rule, but an apparmor one, you will never confine something in `/usr/local/foo` because you don't know how these program have been modified from upstream)

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+    include if exists <local/mutt_editor>
+  }
+  
+  profile pager {
+    include <abstractions/base>
+    include <abstractions/consoles>
+  
+    /{usr/,}bin/less mrix,
+    /{usr/,}bin/more mrix,
+
```
roddhjav, line 148: It needs `@{bin}/pager mr` too.

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,178 @@
...
+    include if exists <local/mutt_editor>
+  }
+  
+  profile pager {
+    include <abstractions/base>
+    include <abstractions/consoles>
+  
+    /{usr/,}bin/less mrix,
+    /{usr/,}bin/more mrix,
```
roddhjav, line 147: `mr` should be enough`.

apparmor.d/profiles-m-r/mutt
doublez13: ah nice catch

apparmor.d/profiles-m-r/mutt
doublez13: I don't think mutt should necessarily have write access to everything underneath `@{HOME}/.mutt/`
Ah but I see the mistake I'll tighten up the later rules.

doublez13: `owner @{HOME}/.mutt/**` is for mutt's misc config files, so this should be read only.
`owner @{HOME}/.mutt*/{,*}` is used for temporary files created by mutt.

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,184 @@
...
+  
+    /{usr/,}bin/less  mr,
+    /{usr/,}bin/more  mr,
+    /{usr/,}bin/pager mr,
+    
+    owner @{HOME}/.lesshsQ r,
+    owner @{HOME}/.local/state/ r,
+    owner @{HOME}/.local/state/less* rw,
+
+    /usr/share/terminfo/x/xterm-256color r,
```
roddhjav, line 159: Use `/usr/share/terminfo/**` instead (color definition change depending of the terminal)

apparmor.d/profiles-m-r/mutt
```diff
@@ -0,0 +1,181 @@
...
+  owner @{HOME}/.mailcap    r,
+  /etc/{mutt/,}mailcap      r,
+  /usr/{local/,}etc/mailcap r,
+
+  # Used to exec programs defined in the mailcap.
+  # There are countless programs that can be executed from the mailcap.
+  # This profile includes only the most basic.
+  @{bin}/{,ba,da}sh  rix,
+ 
+  /{usr/,}{s,}bin/sendmail rPUx,
```
roddhjav, line 83: Use `@{bin}` here too (same elsewhere)

apparmor.d/profiles-m-r/mutt
```diff
@@ -24,6 +24,42 @@ profile mutt @{exec_path} {
 
   @{exec_path} mr,
 
+  # Used to exec programs defined in the mailcap.
+  # There are countless programs that can be executed from the mailcap.
+  # This profile includes only the most basic.
+  @{bin}/{,ba,da}sh               rix,
+ 
+  @{bin}/sendmail                rPUx,
+  /usr/libexec/sendmail/sendmail rPUx,
```
roddhjav, line 33: You have `@{lib}` for `/usr/libexec` (the exact path depends of the distribution)

doublez13: Ah nice! I should have checked the abstractions.

doublez13:
The formatting has been adjusted. Let me know if this meets the development guidelines.
There are some blocks like the following:
```
# Mutt mailcap search path
/etc/{mutt/,}mailcap   r,
/usr/etc/mailcap       r,
owner @{HOME}/.mailcap r,
```
but I assume these are fine as they would fall under similar purpose rules?

doublez13:
Added profile for address book (abook)

roddhjav:
Thanks, this looks better now. Merged!

PR #289, opened 2024-02-03
Systemd-service profile transition

monsieuremre:
Name of the profile changed, so these transitions should also change, no?


roddhjav:
Good catch, merged!

PR #291, opened 2024-02-06
Add Profiles for imv and zathura

valoq:
Both profiles include permissions to read all user files that do not start with a dot in $HOME
Additionally, read permissions are included for /mnt and /media as this would include any possible mount points that follow the reference.

The existing user-read abstraction only includes predefined directories and would not allow access to arbitrary directories in $HOME

### DIFF
--- a/apparmor.d/groups/apps/imv-wayland
+++ b/apparmor.d/groups/apps/imv-wayland
@@ -13,6 +13,7 @@ profile imv @{exec_path} {
   include <abstractions/fontconfig-cache-read>
   include <abstractions/graphics>
   include <abstractions/nameservice-strict>
+  include <abstractions/user-read>
 
   @{exec_path} mr,
 
@@ -22,11 +23,7 @@ profile imv @{exec_path} {
 
   owner @{user_config_dirs}/imv/config r,
 
-  owner @{HOME}/ r,
-  owner @{HOME}/[^.]**  r,
-
-  owner /mnt/** r,
-  owner /media/** r,
+  owner @{MOUNTDIRS}/{,**} r,
 
   owner @{run}/user/@{uid}/imv-*.sock w,
 
--- a/apparmor.d/groups/apps/imv-wayland
+++ b/apparmor.d/groups/apps/imv-wayland
@@ -1,0 +1,34 @@
+# apparmor.d - Full set of apparmor profiles
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/imv-wayland
+profile imv @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/freedesktop.org>
+  include <abstractions/fonts>
+  include <abstractions/fontconfig-cache-read>
+  include <abstractions/graphics>
+  include <abstractions/nameservice-strict>
+
+  @{exec_path} mr,
+
+  /etc/imv_config r,
+  /usr/share/X11/xkb/** r,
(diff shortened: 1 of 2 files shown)


roddhjav:
Hi, Thanks for your PR.

These are images and pdf viewer, they have no legitimate access to the full user directory.

For image viewer, something like the following is already used in the project:
```
  include <abstractions/user-download-strict>
  owner @{user_pictures_dirs}/** r,
```

User with multiple pictures directory can simply edit `@{XDG_PICTURES_DIR}`

For PDF reader, similar to `apparmor.d/profiles-a-f/evince`, you can have:
```
  include <abstractions/user-download-strict>
  include <abstractions/user-read>
  include <abstractions/user-write>
```
It will be easier to maintain the day I decide to restrict it by making configurable by a user.

>  Additionally, read permissions are included for /mnt and /media as this would include any possible mount points that follow the reference.

This is already handled with the `@{user_*_dirs}` variables.

apparmor.d/groups/apps/imv-wayland
```diff
@@ -0,0 +1,34 @@
...
+  include <abstractions/graphics>
+  include <abstractions/nameservice-strict>
+
+  @{exec_path} mr,
+
+  /etc/imv_config r,
+  /usr/share/X11/xkb/** r,
+  /tmp/ r,
+
+  owner @{user_config_dirs}/imv/config
```
roddhjav, line 23: `owner @{user_config_dirs}/imv/config r,`

apparmor.d/groups/apps/imv-wayland
```diff
@@ -0,0 +1,34 @@
...
+
+  owner @{HOME}/ r,
+  owner @{HOME}/[^.]**  r,
+
+  owner /mnt/** r,
+  owner /media/** r,
+
+  owner @{run}/user/*/imv-*.sock w,
```
roddhjav, line 31: That should be `owner @{run}/user/@{uid}/imv-*.sock w,`

apparmor.d/groups/apps/zathura
```diff
@@ -0,0 +1,35 @@
...
+  include <abstractions/base>
+  include <abstractions/freedesktop.org>
+  include <abstractions/fonts>
+  include <abstractions/fontconfig-cache-read>
+  include <abstractions/nameservice-strict>
+  include <abstractions/dconf-write>
+
+  @{exec_path} mr,
+
+  /usr/share/{,**} r,
```
roddhjav, line 19: Do not allow everything in `/usr/share/`. Only needed files/folder.

valoq:
> These are images and pdf viewer, they have no legitimate access to the full user directory.

They have access to all non hidden files and I would argue that these types of applications have indeed legitimate access to the rest of the user directories. Without these rules (or with the user-read abstraction) it would not be possible to view images or PDF in say `/home/myuser/myproject` because that directory has never been specified. 
If I read the project rules correctly, the profile rules should not break normal use usage of the applications. Accessing files in arbitrary home directories is pretty much normal usage for most applications.
I'm not saying restricting the normal usage cannot make sense in some cases but restricting read access to predefined sub directories seems quite random and would contradict rule 2 of the project.

valoq:
I went ahead and changed the whitelisted home directories to the user-read abstraction as suggested.

Regarding the issue of arbitrary directories mentioned above, I created #292 since it is not strictly related to these profiles alone.

valoq:
@roddhjav Is there anything still missing for this to be merged?

roddhjav:
Sorry, kind of forgot about it. It looks good now. Merged! Thanks.

PR #292, opened 2024-02-06
Add new read abstraction

valoq:
This adds a new abstraction that permits safe access to user files in $HOME and typical mount points that follow the filesystem hierarchy standard reference. Dot files in $Home are not included.

Unlike the previous user-read abstraction, this rule set allows access to any (non hidden) directory in $HOME as most application would expect. It thereby follows the defined rule  `2 - Do not break a program` since normal workflow for many apps would include arbitrary directories in $HOME

Note that this PR does replace the previous user-read abstraction with these new rules to fix the workflow of exiting profiles.
The previous user-read is now available as user-xdg-read since it only permits access to predefined XDG directories.


roddhjav:
This is an absolute no.

The project tries to define security boundaries that specifies what objects (files, devices...) can be used by which application. 
The aim is to obtain an architecture that ensures apps and services are isolated without any known ways to escape confinement.

To achieve this goal, we need to ensure that all data folders used by app must be well defined by the project. This is mostly done through the [extended XDG like variables](https://apparmor.pujol.io/variables/). Therefore all data directory accessible under a profile **must** be named under one (or more) of these variables. Hence the actual user-read abstraction for app that can have data almost everywhere.

It may require [some configuration](https://apparmor.pujol.io/configuration/#personal-directories) on the admin side. However, as we follow the xdg definition very closely, most sane default should work out of the box.

valoq:
Allowing access to non hidden user files in $HOME would not allow any confinement escape I am aware of. Whether an application can read the defined XDG directories or any directory in $HOME makes no difference to that objective. Please provide an example if this assumption does not hold up.

Defining specific directories for access makes the confinement a little more similar to the Android model but without a way to share files in a similar way as with Android, this makes things quite complicated for users without adding any security benefit I can see.

Restricting the application in their normal operation can make sense where it actually prevents confinement escape or creates a useful boundary and there are probably very few users who expect their browser to write to places other then `Downloads`, yet is still contradicts the defined project objective mentioned above to some degree, which makes it difficult to access what is intended behavior and what is not.

But the most relevant point is probably what you stated in your linked presentations about users expecting apparmor not getting in their way. If users cannot read an image because it is not located in their specified image directory but e.g. in Documents or an arbitrary directory instead, that is exactly the kind of problem that makes users say "OMG apparmor again". If the apparmor profile restricts the application further then the default security model, it should at least provide a relevant security benefit AND not get in the way of expected behavior. Restricting user-read to predefined directories does not provide any specific security boundary but it does get in the users way.

valoq:
This abstraction is now completely optional and not used by any default profiles.
It can be used to allow users to easily and safely set their read permissions to a more permissive user-read alternative.

This avoids the need for user to define their own more permissive access rules that may accidentally include critical hidden files in $HOME

Unless there is actually any reason not to use this, I would still argue that this abstraction should be the default user-read to improve usability. For now this is only optional.

valoq:
Moved the discussion here: https://github.com/roddhjav/apparmor.d/discussions/307

Note this MR currently only adds an optional abstraction that isn't used by any profile by default as of now.

valoq:
Implemented by https://github.com/roddhjav/apparmor.d/commit/2fc2394bad4f89dc0e367b8bc45c63196e4ef259

PR #293, opened 2024-02-08
Fix systemd-user

monsieuremre:
Fixes the issue.


monsieuremre:
The installation script thing has to be also updated to not look for the drop in file, because it is deleted.

roddhjav:
Thanks, merged. No need to update `make`, it simply install everything that is in `systemd/full/`

PR #298, opened 2024-02-19
Abook: Fix missing directory

doublez13:
Allow abook to create ~/.abook


roddhjav:
Good point. Merged.

Note: This could possibly be simplified as owner `@{HOME}/.abook/{,**} rw,`

PR #315, opened 2024-04-25
Modifications for qbittorrent profile and qt5.d abstractions

yukiteruamano:
This modifications allow read system and user qt5ct configs for better integrations with other DEs (not-KDE).

Reviewing, I see that there are a series of profiles where these changes can be applied (using abstractions/qt5), which allows us to eliminate a couple of lines from the profiles and in this way they integrate better visually when the profile is activated.

The same could be done with Qt6 and qt6ct, although in Debian Stable, at the moment this has very little interest given that most tools are built using Qt5.

If the change is accepted and there are no problems with making the change in the rest of the profiles, I will proceed with it and send the PR.

### DIFF
--- a/apparmor.d/abstractions/qt5.d/complete
+++ b/apparmor.d/abstractions/qt5.d/complete
@@ -4,3 +4,7 @@
 
   /usr/share/qt{,5,6}/qtlogging.ini r,
   /usr/share/qt{,5,6}/translations/*.qm r,
+
+  # Qt5CT
+  /usr/share/qt5ct/{,**} r,
+  owner @{HOME}/.config/qt5ct/{,**} r,
--- a/apparmor.d/abstractions/qt5.d/complete
+++ b/apparmor.d/abstractions/qt5.d/complete
@@ -7,4 +7,4 @@
 
   # Qt5CT
   /usr/share/qt5ct/{,**} r,
-  owner @{HOME}/.config/qt5ct/{,**} r,
+  owner @{user_config_dirs}/.config/qt5ct/{,**} r,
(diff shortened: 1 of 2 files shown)


apparmor.d/abstractions/qt5.d/complete
```diff
@@ -4,3 +4,7 @@
 
   /usr/share/qt{,5,6}/qtlogging.ini r,
   /usr/share/qt{,5,6}/translations/*.qm r,
+
+  # Qt5CT
+  /usr/share/qt5ct/{,**} r,
+  owner @{HOME}/.config/qt5ct/{,**} r,
```
roddhjav, line 10: Cam you use `@{user_config_dirs}` instead of `@{HOME}/.config`

roddhjav:
Thanks for this PR. This is a nice proposal.

> The same could be done with Qt6 and qt6ct, although

Simply usinng `qt{5,6}`ct in the existing abstraction is easier. They is already a similar construction for `qt{5,6}`.

yukiteruamano:
> Simply usinng `qt{5,6}ct` in the existing abstraction is easier. They is already a similar construction for `qt{5,6}`.

That's right, that way, a common base would be created to support those PATHs for Qt5 and Qt6, and the integration of the profiles would be more homogeneous.

The changes that you had mentioned have already been made, if you see no problem with the proposed change, I can start working with the rest of the profiles.

roddhjav:
I meant owner `@{user_config_dirs}/qt5ct/{,**} r,` not `owner @{user_config_dirs}/.config/qt5ct/{,**} r,`

> if you see no problem with the proposed change, I can start working with the rest of the profiles.

Go ahead, makes the PR...

yukiteruamano:
> I meant `owner @{user_config_dirs}/qt5ct/{,**} r,` not `owner @{user_config_dirs}/.config/qt5ct/{,**} r,`
> 
> >

Fix this, sorry!

roddhjav:
Merged, thanks.

PR #318, opened 2024-04-28
Fix access to /tmp using libpam-tmpdir in Debian for qBittorrent

yukiteruamano:
In Debian with the use of libpam-tmpdir, the paths for $TMP and $TMPDIR for PAM sessions are affected by much stronger rules and permissions, providing additional security to the environment.

Those rules for the directory

`/tmp/user/@{uid}/<affected_program>
`

In the case of qBitorrent this applies to the following directory:

`/tmp/user/@{uid}/.qBitorrent
`

This PR fixes the bug and allows qBittorrent to work correctly under these conditions.

Note: This PR would also have positive effects on Whonix, which uses libpam-tmpdir according to this link
(https://forums.whonix.org/t/make-symlink-attacks-and-other-tmp-based-attacks-harder-or-impossible-using-libpam-tmpdir/8488)

### DIFF
--- a/apparmor.d/profiles-m-r/qbittorrent
+++ b/apparmor.d/profiles-m-r/qbittorrent
@@ -108,6 +108,8 @@ profile qbittorrent @{exec_path} {
   owner /tmp/qtsingleapp-qBitto-* rw,
   owner /tmp/qtsingleapp-qBitto-*-lockfile rwk,
   owner /tmp/tmp* rw,
+  owner /tmp/user/@{uid}/.qBittorrent/ rw,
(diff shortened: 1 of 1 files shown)


roddhjav:
Good point, I will have to take a look at this more globally too.
Merged, thanks.

yukiteruamano:
> Good point, I will have to take a look at this more globally too. Merged, thanks.

Yes, it's good to see this globally, first, because I just saw the same issue with the Chromium profile, and I've already patched it (I've also created a profile for Edge, and I'm testing it before submitting the PR).

Second, in the past I have also seen the same issue, in this case in Firefox, when I submitted a fix for this situation with that browser.

PR #324, opened 2024-05-05
plasmashell: add local wallpaper rules

easyteacher:
Allow plasmashell to access wallpapers in the cache folder and the user share folder.


roddhjav:
Thanks. Merged.

PR #326, opened 2024-05-06
Fix access to OpenSC configuration

yukiteruamano:
Hi! 

These changes fix access to OpenSC config in Debian and derivates. In this case, the correct path is `/etc/opensc/opensc.conf`. Only read permission.


roddhjav:
Nice, thanks. Merged

PR #327, opened 2024-05-07
Minor fixes for various abstractions/profiles (exec permission and noise reduction)

yukiteruamano:
Hi! Three minor changes in abstractions/profiles

First, the `abstractions/thumbnails-cache-read` and `abstractions/thumbnails-cache-write`, right now don´t have access to `@{user_cache_dirs}/thumbnails/fail` folder.  For this reason, the `gsd-housekeepin` profile is unable to enter these directories and perform the daily cleaning task that it executes on them.

Adding these dirs to the abstractions solves the problem, and at the same time takes advantage of both abstractions having the same content (and their respective permissions). 

Second, `exim4` in Debian Stable, access to `@{PROC}/sys/net/ipv6/conf/all/disable_ipv6` in read mode, for checking connection with IPv6. This change allow this access and reduce noise in log for `exim4`. 

Finally, `anacron` runs `exim4` upon completion of task execution. This is a default behavior in Debian Stable (12) and right now the profile for `anacron` denied this permission. This change allow this exec permission, and allows `anacron` task information to be sent using `exim4` successfully. 

All this change are tested in Debian Stable using enforced rules for all profile mentioned. Not problem, all ok.

### DIFF
--- a/apparmor.d/abstractions/thumbnails-cache-read
+++ b/apparmor.d/abstractions/thumbnails-cache-read
@@ -7,7 +7,7 @@
   owner @{user_cache_dirs}/thumbnails/{fail,*large,normal}/ r,
   owner @{user_cache_dirs}/thumbnails/{fail,*large,normal}/gnome-thumbnail-factory/ r,
   owner @{user_cache_dirs}/thumbnails/{fail,*large,normal}/gnome-thumbnail-factory/*.png r,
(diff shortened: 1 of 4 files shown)


roddhjav:
Thanks, merged!

Nice improvment over the thumbnails abs.

PR #329, opened 2024-05-11
Create editor abstraction

doublez13:
I'm counting seven child profiles named "editor" that all include roughly the same boiler plate policies. 
This PR creates an editor abstraction and converts a couple profiles over to it.

If you're okay with this, I can open another PR converting the remaining child profiles.


roddhjav:
That's a good idea, thanks.

However, you should move the abstraction in to the `abstractions/app` directory, and you need to remove the base abstraction from it.

roddhjav:
Thanks, merged.

PR #332, opened 2024-05-15
Update discord profile

fira959:
Update for the discord profile, tested on debian and arch (without unprivileged user ns)

### DIFF
--- a/apparmor.d/groups/apps/discord
+++ b/apparmor.d/groups/apps/discord
@@ -41,6 +41,9 @@ profile discord @{exec_path} {
   owner @{tmp}/discord.sock rw,
   owner "@{tmp}/Discord Crashes/" rw,
 
+  deny owner @{user_config_dirs}/discord/0.0.54/modules/** w,
+  owner @{user_config_dirs}/discord/0.0.54/modules/** rm,
+
   owner @{run}/user/@{uid}/discord-ipc-@{int} rw,
 
   include if exists <local/discord>
--- a/apparmor.d/groups/apps/discord
+++ b/apparmor.d/groups/apps/discord
@@ -41,7 +41,7 @@ profile discord @{exec_path} {
   owner @{tmp}/discord.sock rw,
   owner "@{tmp}/Discord Crashes/" rw,
 
-  owner @{user_config_dirs}/discord/*/modules/** rm,
+  owner @{config_dirs}/discord/*/modules/** rm,
 
   owner @{run}/user/@{uid}/discord-ipc-@{int} rw,
 
--- a/apparmor.d/groups/apps/discord
+++ b/apparmor.d/groups/apps/discord
@@ -41,7 +41,7 @@ profile discord @{exec_path} {
   owner @{tmp}/discord.sock rw,
   owner "@{tmp}/Discord Crashes/" rw,
 
-  owner @{config_dirs}/discord/@{int}.@{int}.@{int}/modules/** rwm,
+  owner @{user_config_dirs}/discord/*/modules/** rm,
 
   owner @{run}/user/@{uid}/discord-ipc-@{int} rw,
 
--- a/apparmor.d/groups/apps/discord
+++ b/apparmor.d/groups/apps/discord
@@ -51,7 +51,7 @@ profile discord @{exec_path} {
 
   owner @{user_config_dirs}/discord/** rwm,
 
-  @{PROC}/@{pid}/ r,
+  @{PROC}/@{pid}/task/@{tid}/status r,
 
   owner /dev/pts/@{int} rw,
 
--- a/apparmor.d/groups/apps/discord
+++ b/apparmor.d/groups/apps/discord
@@ -32,10 +32,8 @@ profile discord @{exec_path} {
   @{exec_path} mrix,
   @{sh_path}    rix,
 
-  /opt/discord/chrome-sandbox rix,
-  /opt/discord/chrome_crashpad_handler rix,
-  /usr/share/discord/chrome-sandbox rix,
-  /usr/share/discord/chrome_chrashpad_handler rix,
+  @{lib_dirs}/chrome-sandbox rix,
+  @{lib_dirs}/chrome_crashpad_handler rix,
 
   @{open_path}         rPx -> child-open-browsers,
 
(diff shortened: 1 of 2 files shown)


apparmor.d/groups/apps/discord
```diff
@@ -29,6 +32,11 @@ profile discord @{exec_path} {
   @{exec_path} mrix,
   @{sh_path}    rix,
 
+  /opt/discord/chrome-sandbox rix,
+  /opt/discord/chrome_crashpad_handler rix,
+  /usr/share/discord/chrome-sandbox rix,
+  /usr/share/discord/chrome_chrashpad_handler rix,
```
roddhjav, line 38:
```suggestion
  @{lib_dirs}/chrome-sandbox rix,
  @{lib_dirs}/chrome_crashpad_handler rix,
```

apparmor.d/groups/apps/discord
```diff
@@ -19,6 +19,9 @@ profile discord @{exec_path} {
   include <abstractions/common/electron>
   include <abstractions/thumbnails-cache-read>
   include <abstractions/user-download-strict>
+  include <abstractions/desktop>
+  include <abstractions/graphics>
+  include <abstractions/nameservice-strict>
```
roddhjav, line 24: They are already included in `abstractions/common/electron`

apparmor.d/groups/apps/discord-chrome-sandbox
```diff
@@ -16,8 +16,8 @@ include <tunables/global>
 profile discord-chrome-sandbox @{exec_path} {
   include <abstractions/base>
 
-  capability setgid,
-  capability setuid,
+  #capability setgid,
+  #capability setuid,
```
roddhjav, line 20: Actually needed on some system, do not comment them.

apparmor.d/groups/apps/discord
```diff
@@ -40,5 +48,12 @@ profile discord @{exec_path} {
 
   owner @{run}/user/@{uid}/discord-ipc-@{int} rw,
 
+  owner @{user_config_dirs}/discord/** rwm,
+
+  @{PROC}/@{pid}/task/@{tid}/status r,
```
roddhjav, line 53: This is already included in `abstractions/common/electron`

roddhjav, line 53: Both lines are already included in `abstractions/common/electron`

apparmor.d/groups/apps/discord
```diff
@@ -40,5 +48,12 @@ profile discord @{exec_path} {
...
+
+  # not actually required
+  # owner /dev/pts/@{int} rw,
```
roddhjav, line 56: If it is not needed, do not put it here. Otherwise that would be added using the `consoles` abstraction.

fira959: That's why I left it. The logs show the permission to be used but without it the app runs just fine.
I will remove it entirely to make it clean

apparmor.d/groups/apps/discord-chrome-sandbox
fira959: It wasn't required on debian testing (trixie), even though it has `kernel.unprivileged_userns_clone = 1`
I will test this again on stable to make sure, but I suspect it is no longer a requirement, especially since other electron apps do not seem to need this either (or it is missing)

fira959: Could this be related to this feature?  https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces

I was able to use discord without allowing the two mentioned capabilities, even on systems that have `kernel.unprivileged_userns_clone = 1` but I noticed that my arch test systems also have `kernel.unprivileged_userns_apparmor_policy = 1` and `kernel.apparmor_restrict_unprivileged_unconfined = 0`
If I understand the linked article correctly, it seems the later kernel setting allows the use of unprivileged userns in the context of apparmor? I can't find any documentation about these settings unfortunately, but I suspect they influence the required capabilities.

In any case, I did not find any distribution that required `setuid` or `setgid` and aa-log never complained about them missing either. Maybe they are no longer needed after all?

fira959: According to https://gitlab.com/apparmor/apparmor/-/wikis/unprivileged_userns_restriction the mentioned kernel parameters allow apparmor to block the use of unprivileged userns, unless the apparmor profile includes  `allow userns create`. That would explain why current distributions do not require the capabilities, since applications like discord that are confined by apparmor, will automatically fall back to their suid chrome-sandbox startup script. Do I understand this correctly? If so, what is the consequence for these apparmor profiles? Add userns permissions?

roddhjav: The structure of recent electron based app might have changed. Also, most electron based app disable internal sandboxing.

The new userns restriction feature is handled differently, with the new `userns,` rule that is already part of `abstractions/common/electron`.

Let keep it that way, and I will investigate and update things later. Change could be made to the electron abstraction and applied everywhere.

apparmor.d/groups/apps/discord
roddhjav: Just noticed the `m` in `rwm`. This is not in `abstractions/common/electron`.

There is no way we allow read write and memory mapping on all config dirs. Can you double check what files require it?
For example, in spotify, `rwm` is needed for WidevineCdm: https://github.com/roddhjav/apparmor.d/blob/ad960d477b8037005493fa1ca591d61d4f7936d5/apparmor.d/profiles-s-z/spotify#L41-L42

fira959: There seem to be multile mudule files that require this.
The best I could manage is this:

```
deny owner @{user_config_dirs}/discord/0.0.54/modules/** w,
owner @{user_config_dirs}/discord/0.0.54/modules/** rm,
```

fira959: Changes added as requested.

> we can only ensure `rwm` is limited to the minimum. 
Maybe off topic, but isn't this a security issue in discord by modern standards?

roddhjav: Yes it is. Meanwhile it also depends how it is handled by discord. JIT compilation can be safe if they run in a sandbox (like in chromium).

roddhjav:
Thanks. merged!

PR #333, opened 2024-05-16
Convert remaining profiles to editor abstraction

doublez13:
Although I do not use all the profiles here (AKA, have not tested all of these), this PR just abstracts the common editor rules to the editor abstraction.

The only rule that changes is in the pass profile
Pass profile: ` owner @{user_config_dirs}/vim/{,**} rw,`
editor abstr:  ` owner @{user_config_dirs}/vim/{,**} r,`

None of the other editor profiles allowed write access under the `@{user_config_dirs}/vim/` dir, and I don't see why it would be needed.

### DIFF
--- a/apparmor.d/abstractions/app/editor
+++ b/apparmor.d/abstractions/app/editor
@@ -5,8 +5,7 @@
   include <abstractions/nameservice-strict>
 
   @{bin}/sensible-editor        mr,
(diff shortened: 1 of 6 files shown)


roddhjav:
Thanks a lot. Good PR. Merged!

PR #336, opened 2024-05-25
Minor improvements

fira959:
Maybe the addition of @HOME/.mail should probably be move into an abstraction with the alternative common mail dir locations since they are shared by all mail clients

### DIFF
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -126,7 +126,7 @@ profile mutt @{exec_path} {
 
     /usr/share/terminfo/** r,
     /usr/share/file/misc/magic.mgc r,
-    
+
     owner @{HOME}/ r,
     owner @{HOME}/.lesshs* rw,
     owner @{HOME}/.terminfo/@{int}/* r,
--- a/apparmor.d/profiles-m-r/mutt
+++ b/apparmor.d/profiles-m-r/mutt
@@ -59,6 +59,7 @@ profile mutt @{exec_path} {
   /etc/Muttrc r,
   /etc/Muttrc.d/{*,} r,
 
+  owner @{HOME}/.mail/{,**} rwlk,
   owner @{HOME}/.mail_aliases r, # Common location for mail aliases
   owner @{HOME}/.mailcap r,
   owner @{HOME}/.mime.types r,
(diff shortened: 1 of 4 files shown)


roddhjav:
It already exist (see in `apparmor.d/tunables/home.d/apparmor.d`). You may want to edit `XDG_MAIL_DIR` to `.mail`. Then, you can use `@{user_mail_dirs}` in the profiles.

apparmor.d/profiles-m-r/mutt
```diff
@@ -97,7 +98,7 @@ profile mutt @{exec_path} {
     @{bin}/lynx    mrix,
     
     owner @{HOME}/.w3m/* rw,
-
+    owner @{user_mail_dirs} rw,
```
roddhjav, line 101: You need `owner @{user_mail_dirs}/{,**} rw,` (same elsewhere)

doublez13:
For the Mutt profile, You're granting all the the child profiles `rw` access to `owner @{user_mail_dirs}`. Why?
They should not have access to this.

fira959:
> For the mutt profile, You're granting all the the child profiles `rw` access to `owner @{user_mail_dirs}`. Why? They should not have access to this.

the html-renderer subprofile needs read access to files like ~/.mail/inbox and write access to ~/.mail/tmp/

But the other subprofiles actually didn't need any access like this, not sure why I included them. Permissions should be minimal now

doublez13:
Thanks for updating that :)

Have you overridden a setting in you're mutt config or maybe a custom mailcap definition?
I'm still unclear why the `html-renderer` profile would need any access to your mail directory, as mutt and the html renderer programs would normally communicate via the file `/{var/,}tmp/{.,}mutt*`

If this is something nonstandard that you have overridden in your settings, the rules should live in your personal `local/mutt_html-renderer` file.

Or maybe I'm missing something obvious :)

fira959:
> Thanks for updating that :)
> 
> Have you overridden a setting in you mutt config or maybe a custom mailcap definition? I'm still unclear why the `html-renderer` profile would need any access to your mail directory, as mutt and the html renderer programs would normally communicate via the file `/{var/,}tmp/{.,}mutt*`
> 
> If this is something nonstandard that you have overridden in your settings, the rules should live in your personal `local/mutt_html-renderer` file.
> 
> Or maybe I'm missing something obvious :)

You are correct that I am using a setting for the tmp file that points to that directory. It is not the default but if I recall it is a common recommendation to avoid leaking private data. It could therefore be argued to be a custom change even if it is common. I dont see the harm in allowing access to the tmp folder in any @{user_mail_dirs}. @roddhjav Your call

PR #341, opened 2024-05-30
add multiple profiles

valoq:
This adds profiles for:

**atool 
exiftool
highlight**

It also changes **mediainfo** and adds permissions for reading any user files. This is necessary because unlike what the previous profile suggested, mediainfo can parse more file formats aside from audio and video files and is often used by preview scripts for tui applications. The same is true for exiftool and highlight. 

atool also requires write permissions for user files to allow decompression.

### DIFF
--- a/apparmor.d/profiles-a-f/atool
+++ b/apparmor.d/profiles-a-f/atool
@@ -1,6 +1,5 @@
 # apparmor.d - Full set of apparmor profiles
-# Copyright (C) 2019-2021 Mikhail Morfikov
-# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
+# Copyright (C) 2024 valoq <valoq@mailbox.org>
 # SPDX-License-Identifier: GPL-2.0-only
 
 abi <abi/3.0>,
@@ -8,9 +7,10 @@ abi <abi/3.0>,
 include <tunables/global>
 
 @{exec_path} = @{bin}/atool
-profile atool /{,usr/}{,s}bin/atool {
+profile atool @{exec_path} {
   include <abstractions/base>
-  include <abstractions/user-write>
+  include <abstractions/perl>
+  include <abstractions/user-write-strict>
 
   @{exec_path} mr,
 
@@ -45,8 +45,6 @@ profile atool /{,usr/}{,s}bin/atool {
   @{bin}/unzip rix,
   @{bin}/xz rix,
   @{bin}/zip rix,
-
-  /usr/share/perl5/{,**} r,
 
   include if exists <local/atool>
 }
--- a/apparmor.d/profiles-a-f/atool
+++ b/apparmor.d/profiles-a-f/atool
@@ -1,0 +1,52 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2019-2021 Mikhail Morfikov
+# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/atool
+profile atool /{,usr/}{,s}bin/atool {
+  include <abstractions/base>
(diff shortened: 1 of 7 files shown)


apparmor.d/profiles-a-f/exiftool
```diff
@@ -0,0 +1,20 @@
...
+include <tunables/global>
+
+@{exec_path} = @{bin}/vendor_perl/exiftool
+profile exiftool @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/user-read>
+
+  @{exec_path} mr,
+
+  /usr/share/perl5/{,**} r,
```
roddhjav, line 17: Use   `<abstractions/perl>` instead. Same elsewhere

apparmor.d/profiles-a-f/atool
```diff
@@ -0,0 +1,52 @@
...
+# Copyright (C) 2019-2021 Mikhail Morfikov
+# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/atool
+profile atool /{,usr/}{,s}bin/atool {
```
roddhjav, line 11: You need to set the profile attachment to `@{exec_path}`. Same elsewhere.

apparmor.d/profiles-a-f/atool
```diff
@@ -0,0 +1,52 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2019-2021 Mikhail Morfikov
+# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
```
roddhjav, line 3: Remove Mikhail & I, add you.

apparmor.d/profiles-a-f/atool
```diff
@@ -0,0 +1,52 @@
...
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/atool
+profile atool /{,usr/}{,s}bin/atool {
+  include <abstractions/base>
+  include <abstractions/user-write>
```
roddhjav, line 13: Use both `<abstractions/user-write-strict>` and `<abstractions/user-read-strict>` instead. Same elsewhere.

apparmor.d/profiles-g-l/highlight
```diff
@@ -0,0 +1,23 @@
...
+# Copyright (C) 2019-2021 Mikhail Morfikov
+# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/
+profile highlight /{,usr/}{,s}bin/highlight {
```
roddhjav, line 11:
```suggestion
@{exec_path} = @{bin}/highlight
profile highlight @{exec_path}  {
```

apparmor.d/profiles-a-f/atool
valoq: To be clear, this prevents user from extracting archives located in @{HOME}/ or reading any metadata from files directly located there.
If this is intended, ok. But I do not see what this adds in terms of protection while it will likely get in the way of the user.

valoq:
Also moved previously added profiles into the correct directory

apparmor.d/profiles-a-f/atool
roddhjav: As you may remember the security boundaries we try to maintain: https://github.com/roddhjav/apparmor.d/discussions/307#discussioncomment-8944433. Unlabelled folder/file would be a source of problem (as in: we lost control over what is acceded), then we try to ensure that every user data folder/file is labelled in such a way it can be controlled, revoked, and we can prevent/decrease leak between profile.

roddhjav:
Merged! thanks.

PR #353, opened 2024-06-03
Update kmod

REmerald:
Should fix
`aa-log`
```
ALLOWED kmod open owner @{run}/modprobe.d/ comm=modprobe requested_mask=r denied_mask=r
```
`/var/log/audit.log`
```
type=AVC msg=audit(1717430600.947:2040): apparmor="ALLOWED" operation="open" class="file" profile="lspci" name="/run/modprobe.d/" pid=3053 comm="lspci" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0^]FSUID="remorin" OUID="root"
```
`aa-log --rules`
```
profile kmod {
  @{run}/modprobe.d/ r,
}
```

@roddhjav, should I prepend `owner`, `aa-log` mentiones it, but `--rules` doesn't?

### DIFF
--- a/apparmor.d/profiles-g-l/kmod
+++ b/apparmor.d/profiles-g-l/kmod
@@ -34,6 +34,8 @@ profile kmod @{exec_path} flags=(attach_disconnected) {
   @{lib}/modprobe.d/{,*.conf} r,
   @{lib}/modules/*/modules.* rw,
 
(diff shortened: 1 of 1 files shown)


roddhjav:
That is #352

REmerald:
@roddhjav, sorry, it duplicated for some weird reason

Issue #350, opened 2024-06-03
pacman Failed name lookup - disconnected path


REmerald:
`aa-log`
```
ALLOWED pacman file_inherit owner dev/tty2 info="Failed name lookup - disconnected path" comm=pacman requested_mask=wr denied_mask=wr error=-13
```
`/var/log/audit.log`
```
type=AVC msg=audit(1717433295.882:4787): apparmor="ALLOWED" operation="file_inherit" class="file" info="Failed name lookup - disconnected path" error=-13 profile="pacman" name="dev/tty2" pid=11808 comm="pacman" requested_mask="wr" denied_mask="wr" fsuid=1000 ouid=1000^]FSUID="remorin" OUID="remorin"
```

PR #356, opened 2024-06-04
fix(pacman): add `attach_disconnected`

REmerald:
[Should fix](https://bbs.archlinux.org/viewtopic.php?id=245453) #350 
```
$ aa-log -r | grep pacman -A 2
profile pacman flags=(attach_disconnected) {
  owner dev/tty2 rw,  # file_inherit Failed name lookup - disconnected path
}
```

### DIFF
--- a/apparmor.d/groups/pacman/pacman
+++ b/apparmor.d/groups/pacman/pacman
@@ -7,7 +7,7 @@ abi <abi/3.0>,
 include <tunables/global>
 
 @{exec_path} = @{bin}/pacman
-profile pacman @{exec_path} {
+profile pacman @{exec_path} flags=(attach_disconnected) {
   include <abstractions/base>
   include <abstractions/consoles>
   include <abstractions/disks-read>
--- a/dists/flags/arch.flags
+++ b/dists/flags/arch.flags
@@ -1,7 +1,7 @@
 archlinux-keyring-wkd-sync complain
 makepkg complain
 mkinitcpio attach_disconnected,complain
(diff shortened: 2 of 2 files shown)


roddhjav:
Thanks. I wanted to do some test with this. Can you tell me in witch context this log was raised?

curiosityseeker:
`ALLOWED pacman open owner apparmor/.null info="Failed name lookup - disconnected path" comm=pacman requested_mask=wr denied_mask=wr error=-13`
I had added `deny /apparmor/.null rw,` as a silencer.

I found it also necessary to add `attach_disconnected` to the gpg child profile.

roddhjav:
I meant what was your use of pacman?

REmerald:
> I meant what was your use of pacman?

Actually none, I just boot into the system and get spammed with this error in the logs.
```
ALLOWED pacman file_inherit owner dev/tty2 info="Failed name lookup - disconnected path" comm=pacman requested_mask=wr denied_mask=wr error=-13
```

But, I have `waybar` with `waybar-module-pacman-updates-git` installed, it checks for updates after I boot (actually it seems to do it constanly, but I'm not sure), which is likely to be the cause of the error.

curiosityseeker:
> I meant what was your use of pacman?

Normal update with `sudo pacman -Syu`. It might be related to the mentioned `deny` rule.

roddhjav:
Ok, I think some usage of pacman (with the `--sysroot` option) may require it anyway.

Merged.

PR #361, opened 2024-06-06
fix(xdg-document-portal): add `abstractions/nameservice-strict`

REmerald:
Should fix
```
$ aa-log -f 1 xdg-document-portal 
ALLOWED xdg-document-portal open /etc/nsswitch.conf comm=xdg-document-po requested_mask=r denied_mask=r
ALLOWED xdg-document-portal open /etc/passwd comm=xdg-document-po requested_mask=r denied_mask=r
$ aa-log -f 1 -r xdg-document-portal 
profile xdg-document-portal {
  /etc/nsswitch.conf r,
  /etc/passwd r,
}

```


valoq:
This should probably get the nameservice abstraction instead

`include <abstractions/nameservice-strict>`

REmerald:
> This should probably get the nameservice abstraction instead

Done

roddhjav:
Thanks, merged!

PR #362, opened 2024-06-06
complete atool profile

valoq:
Added nameservice-strict to avoid complains about missing read permissions for /etc/passwd etc.


roddhjav:
Thanks, merged!

PR #365, opened 2024-06-06
Add libreoffice profiles

valoq:
There are many features in LibreOffice that likely require more testing


roddhjav:
Hum... there is already a profile for libreoffice: https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/profiles-g-l/libreoffice

valoq:
Guess I should have checked the profile list more often ...

Issue #348, opened 2024-06-02
chrome_crashpad_handler and stacking


roddhjav:
### Discussed in https://github.com/roddhjav/apparmor.d/discussions/347

<div type='discussions-op-text'>

<sup>Originally posted by **curiosityseeker** May 31, 2024</sup>
Similarly as with Firefox, 

`@{lib_dirs}/chrome_crashpad_handler  rpx,`

in `abstractions/app/chromium` is causing an error. And as in `abstractions/app/firefox`  stacking solves the problem:

`@{lib_dirs}/chrome_crashpad_handler  rpx -> brave//&brave-crashpad-handler,`

However, that abstractions should also work for other Chromium-based browsers, so this rule should better be generalized. I'm not sure how to do this. What do you think?</div>

roddhjav:
@curiosityseeker what does not work with: 
```
 @{lib_dirs}/chrome_crashpad_handler  rPx -> @{profile_name}//&@{profile_name}-crashpad-handler,
```

The alternative solution is to add `@{lib_dirs}/chrome_crashpad_handler  rPx -> brave//&brave-crashpad-handler` in the brave profile and to remove the rule in the abstraction.

curiosityseeker:
> @curiosityseeker what does not work with?:
> 
> ```
>  @{lib_dirs}/chrome_crashpad_handler  rpx -> @{profile_name}//&@{profile_name}-crashpad-handler,

`aa-log` reports:
```
profile brave {
  /opt/brave-bin/chrome_crashpad_handler ix,  # profile transition not found
}
```
And starting `/usr/bin/brave` gives:

` [94073:94073:0603/182730.981969:FATAL:spawn_subprocess.cc(237)] posix_spawn /opt/brave-bin/chrome_crashpad_handler: Permission denied (13)`


> ```
> 
> The alternative solution is to add `@{lib_dirs}/chrome_crashpad_handler rPx -> brave//&brave-crashpad-handler` in the brave profile and to remove the rule in the abstraction.

Yes, I think that's the best solution. Btw., it just comes to my mind that there is a similar problem in `abstractions/app/firefox` once other Firefox-based browsers want to use that abstraction.

PR #366, opened 2024-06-06
Chromium based browsers: add stacking for chrashpad handler

curiosityseeker:
Add stacking for *-crashpad-handler for the Chromium-based browsers as discussed in #348. 

Note that I only tested these changes for Brave (on Arch Linux) and not for Chrome nor Chromium as I don't have them installed. But as the architecture of these browsers is identical it should work for them as well.

Note also that Brave works now in enforce mode.


curiosityseeker:
Oops, forgot Msedge and Opera. Will add them tomorrow.

roddhjav:
Thanks, looks good. Merged

PR #369, opened 2024-06-07
add preview tools

valoq:
Adds profiles for apps that are commonly used to generate file previews in tui applications

### DIFF
--- a/apparmor.d/profiles-a-f/elinks
+++ b/apparmor.d/profiles-a-f/elinks
@@ -1,0 +1,21 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2024 valoq <valoq@mailbox.org>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/elinks
+profile elinks @{exec_path} {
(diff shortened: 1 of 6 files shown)


roddhjav:
Excep `ffmpegthumbnailer`, these programs seems to be [programs-to-not-confine](https://apparmor.pujol.io/development/structure/#programs-to-not-confine)

valoq:
> Except `ffmpegthumbnailer`, these programs seems to be [programs-to-not-confine](https://apparmor.pujol.io/development/structure/#programs-to-not-confine)

> These are general tools that in a general context can legitimately access any file in the system.

None of these are in any way general tools that would need to access arbitrary files more then any application like evince or vlc would.

In case of w3m and elinks, these are literally web browsers like firefox.
The others all operate on normal user files that would only be found in the XDG directories and they fill the exact same role as ffmpegthumbnailer does, so I do not see any reason why they would not fulfill the conditions.

Also, these programs (like any program used as a previewer) expose a higher threat then most other applications since they parse files even without the user intentionally opening them. See gnome tracker for an [example](https://www.bleepingcomputer.com/news/security/gnome-linux-systems-exposed-to-rce-attacks-via-file-downloads/)

roddhjav:
Good point, thanks. Merged

PR #374, opened 2024-06-08
Update dev docs

Stoppedpuma:
Rewords, fixes spelling mistakes, and slightly reformats development docs. Also includes a fix for openSUSE capitalisation.

### DIFF
--- a/docs/development/abstractions.md
+++ b/docs/development/abstractions.md
@@ -2,11 +2,11 @@
 title: Abstractions
 ---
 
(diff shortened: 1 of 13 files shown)


Stoppedpuma:
@roddhjav I would like to propose an additional change related to documentation as well which should probably be decided by you since you're the project owner, across the documentation it seems to be a mixture of American English and International English, I'm looking to change this into just International English to keep spelling consistent across the documentation. Is this a change you would be open to?

roddhjav:
Nice, thanks again. Merged.

PR #380, opened 2024-06-09
feat(profiles-a-f): vim syntax support

REmerald:
Add vim modeline instructing the editor to use syntax plugin provided by apparmor.

Continuation of #379 to keep the diff list relatively short.

PR #379, opened 2024-06-09
feat(tunables, abstactions): vim syntax support

REmerald:
Add `# vim: syntax=apparmor` line, so that vim highlights specific elements.
The feature is from `apparmor` package.
```
$ pacman -Qlq apparmor | grep vim
/usr/share/man/man5/apparmor.vim.5.gz
/usr/share/vim/
/usr/share/vim/vimfiles/
/usr/share/vim/vimfiles/syntax/
/usr/share/vim/vimfiles/syntax/apparmor.vim
```
Doing only `tunables` and `abstractions` for now to keep the `Files Changed` diff relatively short.

PR #381, opened 2024-06-09
feat(profiles-g-l): vim syntax support

REmerald:
Add vim modeline instructing the editor to use syntax plugin provided by apparmor.

Continuation of #379 and #380 to keep the diff list relatively short.


roddhjav:
Thanks, merged

PR #384, opened 2024-06-12
Fix signal menu

fira959:
Selecting the context menu crashes the application. The added permissions avoid this.

I have not added this yet, since everything seems to work just fine without it., but there was one additional denied permission in my logs for signal-desktop-chrome-sandbox:

`capability dac_override,`


roddhjav:
Thanks, merged.

PR #388, opened 2024-06-14
Update variables documentation to include games directories

Stoppedpuma:
Adds the missing game directories and reorganises the variables list. Also adds (s) after "Default Value" to make it more clear that you can add multiple values.

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -6,61 +6,64 @@ title: Variables References
 
 ### User directories
 
-| Description | Name | Default Value |
+| Description | Name | Default Value(s) |
 |-------------|:----:|---------------|
+| Books | `@{XDG_BOOKS_DIR}` | `Books` |
 | Desktop | `@{XDG_DESKTOP_DIR}` | `Desktop` |
-| Download | `@{XDG_DOWNLOAD_DIR}` | `Downloads` |
-| Templates | `@{XDG_TEMPLATES_DIR}` | `Templates` |
-| Public | `@{XDG_PUBLICSHARE_DIR}` | `Public` |
+| Disk images | `@{XDG_IMG_DIR}` | `images` |
 | Documents | `@{XDG_DOCUMENTS_DIR}` | `Documents` |
+| Download | `@{XDG_DOWNLOAD_DIR}` | `Downloads` |
 | Music | `@{XDG_MUSIC_DIR}` | `Music` |
 | Pictures | `@{XDG_PICTURES_DIR}` | `Pictures` |
-| Videos | `@{XDG_VIDEOS_DIR}` | `Videos` |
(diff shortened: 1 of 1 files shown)


docs/variables.md
```diff
@@ -6,61 +6,64 @@ title: Variables References
...
 
-| Description | Name | Default Value |
+| Description | Name | Default Value(s) |
 |-------------|:----:|---------------|
+| Books | `@{XDG_BOOKS_DIR}` | `Books` |
 | Desktop | `@{XDG_DESKTOP_DIR}` | `Desktop` |
-| Download | `@{XDG_DOWNLOAD_DIR}` | `Downloads` |
-| Templates | `@{XDG_TEMPLATES_DIR}` | `Templates` |
-| Public | `@{XDG_PUBLICSHARE_DIR}` | `Public` |
+| Disk images | `@{XDG_IMG_DIR}` | `images` |
```
Stoppedpuma, line 13: I just found this after re-reading my changes, was this lowercase intentionally? The rest of the options start with a capital.

Stoppedpuma:
@roddhjav I plan on also reorganising the order of `tunables/` alphabetically in a separate pr but want to ask why user tunables seem to be split between `xdg-users-dirs.d` and `home.d`? Shouldn't user dirs be put together?

roddhjav:
Thanks,

Some of these variables are sorted by values (`.local/bin` instead of `@{XDG_LIB_DIR}`) can you keep it this way?

> Shouldn't user dirs be put together?

That is to ensure we only use/append already defined variable. See https://github.com/roddhjav/apparmor.d/issues/131
Do **not** move variable defined in a file in another file, it **will** break profile compilation.

Stoppedpuma:
> Some of these variables are sorted by values (.local/bin instead of @{XDG_LIB_DIR}) can you keep it this way?

Just to clarify, this is about the alphabetical sorting for the variables doc? I can undo this if you would like, this was more of a personal nit-pick since "value" sorted ones like "User directories" just look like a mess to my eyes:

image

> That is to ensure we only use/append already defined variable. See https://github.com/roddhjav/apparmor.d/issues/131 Do not move variable defined in a file in another file, it will break profile compilation.

Thanks for clarifying, I wasn't planning on changing the variables between files just wanted to know the reason they were split. If you want the sorting I did undone as stated above I won't open a PR since there's no reason to.

Stoppedpuma:
I missed a few other variables while reading through tunables, don't merge this quite yet

docs/variables.md
roddhjav: Yes, it was as it is not really a "media" like file.

roddhjav:
Thanks, merged.

docs/variables.md
nobody43: As this is not straightforward, maybe we should change the directory name to more verbose one.

PR #390, opened 2024-06-15
feat(profiles-m-r): vim syntax support

REmerald:
Add vim modeline instructing the editor to use the syntax plugin provided by apparmor.

Continuation of #379, #380 and #381 to keep the diff list relatively short.


roddhjav:
Thanks, merged

PR #391, opened 2024-06-15
feat(profiles-s-z): vim syntax support

REmerald:
Add vim modeline instructing the editor to use the syntax plugin provided by apparmor.

Continuation of #379, #380, #381, #390 to keep the diff list relatively short.


roddhjav:
thanks, merged.

PR #396, opened 2024-06-16
feat(groups/{u,v,w,x}*): vim syntax support

REmerald:
Add vim modeline instructing the editor to use syntax plugin provided by apparmor.
Continuation of #395 to keep the diff list relatively short.


roddhjav:
Thanks, they have all been merged.

PR #408, opened 2024-07-11
add profiles for cmus and ouch

valoq:
Ouch is a command-line (de)compression tool written in rust

cmus is a lightweight ncurses music player


roddhjav:
Thanks, merged.

PR #419, opened 2024-07-17
Fix hyprlock

odomingao:
Small fix. I'd like to take this opportunity to ask whether it's fine to include this rule:

` owner /dev/shm/@{uuid} r,` 

in other profiles, as this seems to be required for many profiles to work with the currently WIP Hyprland profile. Currently, other than hyprlock, the profiles that I needed to add this are: keepassxc, mullvad-gui, pavucontrol, strawberry, xwayland.


roddhjav:
I think `owner /dev/shm/@{uuid} r,` should go into https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/abstractions/wayland.d/complete instead.

PR #423, opened 2024-07-21
add hyprland profile

odomingao:


AveraheLinuxUser:
Hello, every gui app needs acess to read /dev/shm/@{uuid} as owner, which created by hyprland after launch.
It can be add as new abstraction called hyprland, or added to exists desktop abstraction

/etc/apparmor.d/abstractions/hyprland

# Copyright (C) 2022-2024 Alexandre Pujol <alexandre@pujol.io>

  owner /dev/shm/@{uuid} r,

# vim:syntax=apparmor

odomingao:
> Hello, allmost every gui app needs acess to read /dev/shm/@{uuid} as owner, which created by hyprland after launch. It can be add as new abstraction called hyprland, or added to exists desktop abstraction
> 
> /etc/apparmor.d/abstractions/hyprland
> 
> owner /dev/shm/@{uuid} r,

Yes, the desktop abstraction probably makes the most sense but there are graphical applications that don't need the abstraction. I think it should be just `/dev/shm/@{uuid} r,` rather than `owner /dev/shm/@{uuid} r,`, because graphical applications can be run as other users depending on each individual's setup.

roddhjav:
`owner /dev/shm/@{uuid} r,` us required for wayland, thefore it needs to be added to the wayland abstraction, not in the desktop abs.

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/Hyprland
+profile hyprland @{exec_path} flags=(attach_disconnected) {
+  include <abstractions/base>
+  include <abstractions/desktop>
+  include <abstractions/graphics>
+  include <abstractions/wayland>
```
roddhjav, line 14: the wayland abs is already included in the desktop abs. Therefore, you can safely remove it here.

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+
+  ptrace read,
+
+  network netlink raw,
+
+  signal send,
+
+  @{exec_path} mr,
+
+  @{bin}/** rPUx,
```
roddhjav, line 26: Use the app-launcher-user abstraction instead of this line.

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+  @{user_bin_dirs}/** rPUx,
+  owner @{user_share_dirs}/hyprpm/** mr,
+
+  /usr/share/hyprland/{,*} r,
+  /usr/share/libinput/{,*} r,
+
+  owner @{user_cache_dirs}/hyprland/** w,
```
roddhjav, line 33: It is safe, and time saving to directly allow `rw` on its own cache.

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+  owner @{user_config_dirs}/hypr/hyprland.conf r,
```
roddhjav, line 34: Same for config, it is it own config file:
```
owner @{user_config_dirs}/hypr/** r,
```

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+  @{run}/udev/data/+dmi:id r,
+  @{run}/udev/data/+drm:card@{int}-* r,
+  @{run}/udev/data/+input:input@{int} r,
+  @{run}/udev/data/+pci:* r,
+  @{run}/udev/data/+platform:pcspkr r,
+  @{run}/udev/data/+sound:card@{int} r,
+  @{run}/udev/data/+usb:* r,
+  @{run}/udev/data/c13:@{int} r,
+  @{run}/udev/data/c189:@{int} r,
+  @{run}/udev/data/c226:@{int} r,
```
roddhjav, line 47: It is easier to keep (and copy paste) existing structure already used around the profiles:

```suggestion
  @{run}/systemd/sessions/@{int} r,
 
  @{run}/udev/data/+acpi:* r,             # for acpi
  @{run}/udev/data/+dmi:id r,             # for motherboard info
  @{run}/udev/data/+drm:card@{int}-* r,   # For screen outputs
  @{run}/udev/data/+input:input@{int} r,  # for mouse, keyboard, touchpad
  @{run}/udev/data/+pci:* r,              # Identifies all PCI devices (CPU, GPU, Network, Disks, USB, etc.)
  @{run}/udev/data/+platform:* r,
  @{run}/udev/data/+sound:card@{int} r,   # for sound card
  @{run}/udev/data/+usb* r,               # for USB mouse and keyboard
  @{run}/udev/data/c13:@{int} r,          # for /dev/input/*
  @{run}/udev/data/c189:@{int}  r,        # for /dev/bus/usb/**
  @{run}/udev/data/c226:@{int} r,         # for /dev/dri/card* 
```

Also, please move all `@{run}/udev/data` rules after the temp directories as they are quite different (`@{run}/user/@{uid}`, `/tmp/`, `/dev/shm`)

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+  owner @{run}/user/@{uid}/gamescope-* rw,
+  owner @{run}/user/@{uid}/hypr/{,**} rw,
+  owner @{run}/user/@{uid}/.hyprpaper_* rw,
+  owner @{run}/user/@{uid}/.hyprpicker_* rw,
+  owner /tmp/.X@{int}-lock w,
+  owner /dev/shm/@{uuid} rw,
```
roddhjav, line 53: Move this to the wayland complete file.

apparmor.d/groups/hyprland/hyprland
```diff
@@ -0,0 +1,71 @@
...
+  owner /dev/shm/.org.chromium.Chromium.@{rand6} rw,
+
+  @{sys}/bus/ r,
+  @{sys}/class/input/ r,
+  @{sys}/devices/@{pci}/oot_vga r,
```
roddhjav, line 58: oot_vga does not exist. It should be `@{sys}/devices/@{pci}/boot_vga r,`

apparmor.d/groups/hyprland/hyprland
odomingao: is it ok to remove `owner` from  `owner /dev/shm/@{uuid} rw,`? Applications running as a different user will not launch otherwise

roddhjav: I would say no, (simply add a rule without owner in you local addition). This kind of runtime specific settings does not fit well with generic apparmor profiles. This kind would better be handled directly by the program that handle starting the app as a different user.

odomingao:
I think that I fixed everything you mentioned. This: `owner /dev/shm/grim-@{rand6} rw,` is needed for XDPH. I added it here because I thought it makes sense to be in the wayland abstraction, but it's not needed for the Hyprland profile itself

roddhjav:
Thanks. It looks good now. Merged

PR #433, opened 2024-08-04
Fix mpv (ao=alsa)

odomingao:
mpv needs access to /dev/snd files for the alsa audio backend to work


roddhjav:
This should be handled by the audio-server abstraction. It should be fixed now.

PR #449, opened 2024-08-25
Update systemd-journald

odomingao:
apparmor="DENIED" operation="open" class="file" profile="systemd-journald" name="/run/udev/data/+mdio_bus:r8169-0-300:00"  comm="systemd-journal" requested_mask="r" denied_mask="r" fsuid=0 ouid=0 FSUID="root" OUID="root"


roddhjav:
Thanks, merged.

PR #454, opened 2024-08-30
socat: add apparmor profile

nishitm:
Creating a profile for "socat". I have tested this profile on "socat" latest upstream version: 1.8.0.1 with its test suite on kernel `6.8.0-31`. I have written this profile with the latest abi 4.0 support. 

Let me know if it fails any tests or needs more modifications. Thanks

### DIFF
--- a/apparmor.d/groups/network/socat
+++ b/apparmor.d/groups/network/socat
@@ -1,20 +1,18 @@
-#------------------------------------------------------------------
-#    Author: Nishit Majithia (nishitm)
-#
-#    This program is free software; you can redistribute it and/or
-#    modify it under the terms of version 2 of the GNU General Public
-#    License published by the Free Software Foundation.
-#    SPDX-License-Identifier: GPL-2.0-only
-#------------------------------------------------------------------
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2024 Nishit Majithia (nishitm)
+# SPDX-License-Identifier: GPL-2.0-only
 # vim: ft=apparmor
 
-abi <abi/4.0>,
+abi <abi/3.0>,
 
 include <tunables/global>
 
-profile socat /usr/bin/socat {
+@{exec_path} = /usr/bin/socat
+profile socat @{exec_path} {
   include <abstractions/base>
   include <abstractions/nameservice-strict>
+  include <abstractions/ssl_certs>
+  include <abstractions/consoles>
 
   capability dac_read_search,
   capability dac_override,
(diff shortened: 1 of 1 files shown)


apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,59 @@
...
+  file rw /dev/tty,
+
+  # Allow reading /dev/vsock
+  file r /dev/vsock,
+
+  # certs/keys can be are stored in:
+  #     - /etc/ssl/certs/*.{key,crt}
+  #     - $HOME/.cert/**/*.pem
+  file r /etc/ssl/certs/{,*.{key,crt}},
+  file r @{HOME}/.certs/{,**},
```
roddhjav, line 55: Use the `ssl_certs` abstraction instead.

roddhjav:
Thanks for this PR.

I would encourage you to read the profile guideline: https://apparmor.pujol.io/development/#add-a-profile and to adapt the profile accordingly.

```
  # Allow reading from /dev/tty
  file rw /dev/tty,
```
While it is a valid apparmor rule, we use the more common form:   `/dev/tty rw,` (btw, this comment is useless).

> I have written this profile with the latest abi 4.0 support.

You don't use abi 4 specific feature, so keep it in abi 3. I will all profiles at once when we move to abi 4 by default. (ABI 4 is already supported in distribution that support it).

apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,59 @@
...
+  capability net_bind_service,
+  capability sys_resource,
+
+  # Allow creation of network sockets and `socat` uses dccp for some
+  # fuctionalities that is why it is necessary to allow whole `network`
+  network,
+
+  # Enale /dev/ptmx access for testsuite
+  # file rw /dev/ptmx,
+  # file rw /dev/pts/*,
```
roddhjav, line 36: Use the `consoles` abstraction. (same for /dev/tty).

apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,59 @@
...
+
+  # TUN/TAP device
+  file rw /dev/net/tun,
+
+  # Process-specific access
+  file rw @{PROC}/@{pid}/fd/*,
```
roddhjav, line 42: Should be: `@{PROC}/@{pid}/fdinfo/@{int} rw,`

nishitm:
thanks for the valuable feedback, I have updated the profile in https://github.com/roddhjav/apparmor.d/pull/454/commits/b9eb78333806580e15b3b4c2252c7b4887082e22

apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,47 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2024 Nishit Majithia (nishitm)
+# SPDX-License-Identifier: GPL-2.0-only
+# vim: ft=apparmor
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = /usr/bin/socat
```
roddhjav, line 10: It must be: `@{bin}/socat`

apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,47 @@
...
+  capability sys_admin,
+  capability fsetid,
+  capability chown,
+  capability net_bind_service,
+  capability sys_resource,
+
+  # Allow creation of network sockets and `socat` uses dccp for some
+  # fuctionalities that is why it is necessary to allow whole `network`
+  network,
+
```
roddhjav, line 31: You need to insert the rule `@{exec_path} mr,`

nishitm: Change has been added

nishitm:
aah, I missed those pieces 
but it has been added. Thanks for pointing this out

roddhjav:
Good, merged.

apparmor.d/groups/network/socat
```diff
@@ -0,0 +1,49 @@
...
+  # Enale /dev/ptmx access for testsuite
+  # /dev/ptmx rw,
+
+  # TUN/TAP device
+  /dev/net/tun rw,
+
+  # Process-specific access
+  @{PROC}/@{pid}/fdinfo/@{int} rw,
+  @{PROC}/@{pid}/stat           r,
```
nobody43, line 42: @nishitm 
Not possible to restrict with `owner`?

PR #457, opened 2024-09-01
Add support for gnome-boxes

EricLin0509:


roddhjav:
Thanks for this PR. However, I am not going to merge it in its current stage, as they are too many issues with it:

The important thing to consider is that *gnome-boxes* is **only** a nice UI for managing VM (see the profile for virt-manager that does the same), it is not the VM manager itself. Therefore all of the programs that handle some part of the VM should have they own independent profile. Most of them already have one in this project.
To sum up: all `@{bin}/virt*` services and `@{bin}/qemu*` should be in `Px` transition. Because you currently have a profile for `ix`, it means you have a lot of useless rules and it is better to start to profile from scratch again.

A lot of accesses can be simplified by the use of abstractions. Have a look at: `<abstractions/graphics>`, `<abstractions/desktop>`, `<abstractions/gstreamer>`,  `<abstractions/user-download-strict>`, `<abstractions/consoles>`

`@{run}/udev/data/{,**} r,` is too large. See: https://apparmor.pujol.io/development/structure/#udev-rules

The entry point rule is missing: `@{exec_path} mr,`

PR #458, opened 2024-09-01
Update chronyd

odomingao:
apparmor="DENIED" operation="create" class="net" profile="chronyd"  comm="chronyd" family="inet" sock_type="stream" protocol=0 requested_mask="create" denied_mask="create"


roddhjav:
Thanks, merged.

PR #467, opened 2024-09-05
Update profile for ufw

EricLin0509:
Fix issue: https://github.com/roddhjav/apparmor.d/commit/d5ee5c51cba17bea2647e068e90f610e820e6aff#r146255885


nobody43:
If underscores may be present in the paths, you should use `????????` instead of `@{rand8}`.

EricLin0509:
Oh, I see. Thank you so much! @nobody43

roddhjav:
Thanks, merged.

PR #474, opened 2024-09-10
Add support for linuxqq

EricLin0509:


roddhjav:
linuxqq is an electron app, therefore you should use the [electron abstraction](https://apparmor.pujol.io/development/abstractions/#commonelectron) insteead of the chromium one. Also, `bwrap` is not needed here.

Please update the profile accordingly.

EricLin0509:
OK, It should be fixed now

roddhjav:
Thanks, merged

roddhjav:
I was a bit fast to merge, for your future contribution, here is the commit that finish the profile fb93ac0df35346019181d96bee9624925e04fb84:
- Use `@{open_path} rPx -> child-open-strict,` instead of `xdg-open`. The child profile ensures it can only open a limited set of program.
- Use `@{lib_dirs}` inside the profile.
- The abstraction `common/electron` include `graphics`, `ssl_certs` and `nameservice-strict`

PR #475, opened 2024-09-10
Add support for controllers in game abstraction

odomingao:


roddhjav:
These are legacy joystick devices, the newer one use `/dev/input/event@{int}`. However, it is perfectly valid to add them here. Thanks.

PR #489, opened 2024-09-15
add ssh-sk-helper

valoq:
Required for use of ssh with security keys like fido devices

### DIFF
--- a/apparmor.d/groups/ssh/ssh
+++ b/apparmor.d/groups/ssh/ssh
@@ -25,7 +25,7 @@ profile ssh @{exec_path} {
 
   @{bin}/@{shells}     rUx,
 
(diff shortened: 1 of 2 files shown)


apparmor.d/groups/ssh/ssh
```diff
@@ -25,6 +25,8 @@ profile ssh @{exec_path} {
 
   @{bin}/@{shells}     rUx,
 
+  @{lib}/ssh/ssh-sk-helper rix -> ssh-sk-helper,
```
roddhjav, line 28: That should be `@{lib}/ssh/ssh-sk-helper rPx,`

valoq:
This should be fixed now

roddhjav:
Thanks, merged.

PR #512, opened 2024-09-22
Create vesktop

odomingao:


apparmor.d/profiles-s-z/vesktop
```diff
@@ -0,0 +1,46 @@
...
+
+  @{sys}/devices/@{pci}/usb@{int}/**/interface r,
+
+        @{PROC}/@{pid}/cmdline r,
+  owner @{PROC}/@{pid}/task/@{tid}/comm rw,
+
+  owner /dev/ r,
+
+  deny       /dev/tty rw,
+  deny owner /dev/tty@{int} rw,
```
roddhjav, line 41: Why do you deny them?

odomingao: As a silencer. It asks for these permissions but I don't think that it actually needs them, as everything on the program works fine, and AFAIK (correct me if I'm wrong) access to /dev/tty allows the program to escape confinement so I thought it was better to deny.

roddhjav: Do you have any source of this?

odomingao: For the possible /dev/tty escapes? https://www.errno.fr/TTYPushback.html and https://github.com/containers/bubblewrap/issues/142

roddhjav:
Thanks, merged.

apparmor.d/profiles-s-z/vesktop
roddhjav: Thanks. This has been discussed here: https://github.com/roddhjav/apparmor.d/discussions/340

Spoiler alert: the solution is not apparmor.

PR #513, opened 2024-09-23
New profile: protonmail

curiosityseeker:


apparmor.d/profiles-m-r/protonmail
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{name} = "Proton Mail"
+@{lib_dirs} = /opt/@{name}
+@{config_dirs} = @{user_config_dirs}/@{name}
+@{cache_dirs} = @{user_cache_dirs}/@{name}
```
roddhjav, line 12: Name should be: `@{name} = proton-mail "Proton Mail"`

roddhjav:
Thanks for this PR.

Some minor requested changes:

1. You [need](https://github.com/flathub/me.proton.Mail/blob/master/me.proton.Mail.yml)  to give access to `bus/org.freedesktop.secrets`:
    ```
      include <abstractions/bus-session>
      include <abstractions/bus/org.freedesktop.secrets>
    ```
    Thus, you can remove `/etc/machine-id r,`

2.  exec_path need to be set to: `@{exec_path} = @{bin}/proton-mail /opt/proton-mail/Proton*`
Because:
    - `@{bin}/proton-mail` is also a possible entrypoint
    - There is a bug in the pre-built of this package, it does not support quoted space in variable resolution. I will fix it some day, but it is not a priority.

curiosityseeker:
Should be fixed now.

roddhjav:
Thanks, merged.

PR #531, opened 2024-09-26
Create pokemmo

odomingao:
(It's a native online game.)

If this is merged, then compositors also need to include the glfw abstraction, or else the game will not launch (even if this profile is in complain mode).

I've highlighted/separated sections of the profile that are unused by the game itself, in case we ever decide to split the profile. This didn't seem feasible to me at the time of writing because of nnp and because both the installer and the game use some of the same binaries.

### DIFF
--- a/apparmor.d/profiles-m-r/pokemmo
+++ b/apparmor.d/profiles-m-r/pokemmo
@@ -1,0 +1,100 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2024 odomingao
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/pokemmo-launcher
+profile pokemmo @{exec_path} flags=(attach_disconnected) {
+  include <abstractions/base>
+  include <abstractions/audio-client>
+  include <abstractions/dconf-write> # Installer
+  include <abstractions/desktop>
+  include <abstractions/fontconfig-cache-write> # Installer
+  include <abstractions/glfw>
+  include <abstractions/graphics>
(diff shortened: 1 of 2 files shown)


apparmor.d/profiles-m-r/pokemmo
```diff
@@ -0,0 +1,100 @@
...
+  @{bin}/python ix,
+  @{bin}/python@{int} ix,
+  @{bin}/python@{int}.@{int} ix,
+  @{bin}/which ix,
+  @{lib}/jvm/java-@{int}-openjdk/bin/java ix,
+
+  # Installer
+  @{bin}/openssl ix,
+  @{bin}/ps ix,
+  @{bin}/wget ix,
```
roddhjav, line 48: `@{bin}/ps Px,`

apparmor.d/profiles-m-r/pokemmo
```diff
@@ -0,0 +1,100 @@
...
+  @{exec_path} mr,
+
+  @{coreutils_path} ix,
+  @{sh_path} ix,
+
+  @{bin}/java ix,
+  @{bin}/perl ix,
+  @{bin}/python ix,
+  @{bin}/python@{int} ix,
+  @{bin}/python@{int}.@{int} ix,
```
roddhjav, line 41: `@{python_path} ix,`

apparmor.d/profiles-m-r/pokemmo
odomingao: can't due to nnp

apparmor.d/profiles-m-r/pokemmo
odomingao: fixed

roddhjav:
Thanks, merged!

PR #544, opened 2024-10-04
`xeyes`

beroal:


apparmor.d/profiles-s-z/xeyes
```diff
@@ -0,0 +1,19 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2022-2024 Roman Beslik <me@beroal.in.ua>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/3.0>,
```
roddhjav, line 5:
```suggestion
abi <abi/4.0>,
```

beroal: I would like to know how to install AppArmor 4. Otherwise these profiles won't work on my computer.

roddhjav: Don't worry, the build system know witch ABI it needs to target and converts the profiles from 4 to 3 if needed.
See:
* https://github.com/roddhjav/apparmor.d/blob/18a71512a9519585216f6fec11b4b7371630990f/cmd/prebuild/main.go#L38
* https://github.com/roddhjav/apparmor.d/blob/18a71512a9519585216f6fec11b4b7371630990f/pkg/prebuild/builder/abi.go#L13

It is important, as I am currently testing some abi 4 only feature.

roddhjav:
Thanks, merge!

PR #546, opened 2024-10-04
aMule, a file sharing program; not all executables

beroal:


apparmor.d/profiles-a-f/amule
```diff
@@ -0,0 +1,53 @@
...
+  include <abstractions/nameservice-strict>
+
+  network inet dgram,
+  network inet stream,
+  network inet6 dgram,
+  network inet6 stream,
+  network netlink raw,
+
+  @{sh_path} Cx -> shell, 
+
```
roddhjav, line 23: If the shell is part of amule (ie: not interactive, not related to an extension...), the subprofile is useless.

roddhjav:
- There is the `@{user_torrents_dirs}` variable that should probably be used in this context instead of documents.
- Use `abi/4.0`

roddhjav:
Thanks, merged.

PR #551, opened 2024-10-10
New profile: ksystemstats

curiosityseeker:


apparmor.d/groups/kde/ksystemstats
```diff
@@ -0,0 +1,50 @@
...
+
+  /etc/fstab r,
+  /etc/libnl/classid r,
+  /etc/sensors.d/ r,
+  /etc/sensors3.conf r,
+
+  @{run}/udev/data/+drm:card@{int}-* r,
+  @{run}/udev/data/c226:* r,
+  @{run}/udev/data/b259:* r,
+  @{run}/udev/data/b8:@{int} r,
```
roddhjav, line 27: I try to keep uniformity about udev data, and to have inline doc.
```suggestion
  @{run}/udev/data/+drm:card@{int}-* r,   # For screen outputs
  @{run}/udev/data/b8:@{int} r,       # for /dev/sd*
  @{run}/udev/data/c226:@{int} r,         # For /dev/dri/card[0-9]*
  @{run}/udev/data/b259:@{int} r,     # Block Extended Major
```

Note: `udev/data/b8`,  `udev/data/b259` are part of the `disks-*` abs. And are  really link to the disks. That could indicate the disks-read abs is needed. If so, some rule in `@{sys}/` could be restricted as `disks-*` covert a lot.

PR #552, opened 2024-10-10
Update gmenudbusmenuproxy and xembedsniproxy

curiosityseeker:


apparmor.d/groups/kde/gmenudbusmenuproxy
```diff
@@ -25,6 +26,10 @@ profile gmenudbusmenuproxy @{exec_path} {
   owner @{user_config_dirs}/gtk-{2,3}.0/settings.ini{,.@{rand6}} rwl,
   owner @{user_config_dirs}/gtk-{2,3}.0/settings.ini.lock rwk,
 
+  owner @{run}/user/@{uid}/iceauth_@{rand6} r,
```
roddhjav, line 29: It should be already included in `X-strict` (itself included in `kde-strict`)

PR #553, opened 2024-10-13
A fix for fwupdmgr

EricLin0509:
Fix no output issue when running fwupdmgr, here is the log: 
`apparmor="DENIED" operation="file_inherit" class="file" profile="fwupdmgr" name="/dev/pts/1"  comm="fwupdmgr" requested_mask="wr" denied_mask="wr" fsuid=0 ouid=0 FSUID="root" OUID="root"`


roddhjav:
Thanks, merged.

PR #561, opened 2024-10-15
Prepare for re-attached-path

EricLin0509:


roddhjav:
Thanks,

Did you checked these with re-attached-path enabled?

EricLin0509:
Yes, I have tested on my VM with re-attached-path enabled.

PR #563, opened 2024-10-17
YubiKey support for sudo

LordGrimmauld:
This PR is an attempt to support YubiKeys for sudo authentication. Tested on the highly experimental NixOS concoction, but works well there. Logs are clean and the widening allowed here is minimal.

- the yubikey is a u2f usb device, so usb abstraction is required
- the authentication with yubikey against sudo happens as challenge response, which is why rw on the challenge file is required
- the elevator first checks whether a .yubico folder exists, which is why reading the folder (but not the files within) is required

### DIFF
--- a/apparmor.d/abstractions/app/sudo
+++ b/apparmor.d/abstractions/app/sudo
@@ -12,6 +12,7 @@
   include <abstractions/consoles>
   include <abstractions/nameservice-strict>
   include <abstractions/wutmp>
+  include <abstractions/devices-usb>
 
   capability audit_write,
   capability dac_override,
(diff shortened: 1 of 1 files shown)


roddhjav:
Thanks, for this PR.

- Can you move the `@{HOME}` rules next to the other one (`owner @{HOME}/.sudo_as_admin_successful rw,`) in the file.
- Can you double check that `@{HOME}/.yubico/ r,` is required (instead of `owner @{HOME}/.yubico/ r,`

LordGrimmauld:
`type=AVC msg=audit(1729195710.153:3204): apparmor="ALLOWED" operation="open" class="file" profile="sudo" name="/home/grimmauld/.yubico/" pid=178050 comm="sudo" requested_mask="r" denied_mask="r" fsuid=0 ouid=1000FSUID="root" OUID="grimmauld"`
Notice owner uid is 1000 (user) and  fsuid is 0 (root)

roddhjav:
Perfect. Merged.

PR #569, opened 2024-10-22
Update to ABI4.0 for Arch Linux

EricLin0509:
The apparmor-4.0.3-1 package from Arch Linux had been moved out of the testing repository, See [here](https://archlinux.org/packages/extra/x86_64/apparmor/)


roddhjav:
Excellent timing, I was going to work on this myself. Merged.

PR #573, opened 2024-10-22
add startlxqt

Besanon:


roddhjav:
`startlxqt` should be in the group `lxqt`.

Besanon:
set up new directory

Issue #464, opened 2024-09-04
RFC: Split the profiles in multiple packages.


roddhjav:
This issue aims to be an open discussion about possible reorganization of how the profiles are shipped.

The current apparmor.d package ships a lot of profiles. Depending on the target system, they are not always needed or not always stable. It can be an issue for new users as they can be legitimately afraid of breaking their system even if over 90% of the profiles are safe to use.

The idea is to split the profiles in multiple packages. The project itself would remain a mono repository with all the profiles in it, as it make it easier to develop and maintain them.

### Possible packages

**`apparmor.d.base`**

Minimal core with tunables, abstractions, and dependencies of other profiles such that one can use this package if they only want to confine a given profile such as a server app, a GUI app without having to confine the rest of the system.

Possible issues: dbus may not be part of `base` and peer_label should be handled accordingly. 

**`apparmor.d.core`**

Profiles for program installed by default and common to all desktops and servers. Eg:
- coreutils (and similar)
- systemd
- dbus
- grub

All profile in *core* should be safe to be enforced by default (they may be a few exceptions such as udevd).

**`apparmor.d.desktop`**

Base desktop, also useful for unsupported DE. Equivalent of the following groups:
- `freedesktop`
- `gvfs`

**`apparmor.d.gnome`, `apparmor.d.kde`, `apparmor.d.xfce`**

Desktop specific package. Should conflict with each other. DE specific configuration (in gnome) may require us to have a conflict with `apparmor.d.desktop` too.

**`apparmor.d.browsers`**

All profiles related to browsers.

**`apparmor.d.app`**

Profiles related to UI applications. Could be split in multiple sub-packages.

Ideally, this package should only have `base` as strict dependency. However, it may lead to issue.

**`apparmor.d.full`**

Like the current `apparmor.d`

### Advantages

- Provide a core set of battle tested profiles that can safely be enforced by default.
- Only provide the minimal set of profiles required for a specific use case:
    - no more profiles for desktop on a server,
    - no profile for KDE on Gnome.
- DE specific package can only support one DE and ship DE specific configuration.
- A user using a DE not supported could still use the core package, the app package, and the browser package.
- In FSP mode, not having a profile means preventing the program from running.

### Disadvantages

- Need to sort all profiles into groups/area to determine which profiles go where
- Need to ensure that profile dependencies are correctly set
- Need to ensure each possible combination of installed packages work.
- Lot of testing and possible maintenance.
- Breaking change would require manual update from current set-up

[...]

PR #578, opened 2024-10-23
DRAFT: Split the profiles in multiple packages.

roddhjav:
This PR provide early support for multiple packages `apparmor.d.*`. See #464 for more context.

Example of packages generated:
- `apparmor.d` The same than usual, no change.
- `apparmor.d.base` 
- `apparmor.d.other` All other profiles that are part of this project and not installed by other `apparmor.d.*`


roddhjav:
The final architecture will be different; so this is not needed anymore. See #986

PR #580, opened 2024-10-24
Add lxqt-session, enable start in sddm

Besanon:


apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+  include <abstractions/qt5-shader-cache>
+  include <abstractions/nameservice-strict>
+
+  signal (send),
+  signal (receive) set=(kill, term) peer=startlxqt,
+  signal (receive) set=(kill, term) peer=sddm,
+
+  ptrace (read),
+
+  network netlink raw,
```
roddhjav, line 24: This rule should come before the signal rules.

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/lxqt-session
+profile lxqt-session @{exec_path} flags=(attach_disconnected) {
+  include <abstractions/base>
+  include <abstractions/dbus-accessibility>
```
roddhjav, line 13: Use `bus-accessibility` instead (see https://apparmor.pujol.io/development/dbus/)

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+  @{bin}/lxqt-leave                   rPx,
+  @{bin}/lxqt-about                   rPx,
+  @{bin}/lxqt-config-monitor          rPx,
+  @{bin}/dbus-update-activation-environment rCx -> dbus,
+  @{bin}/systemctl                    rCx -> systemctl,
+
+  @{bin}/pavucontrol                  rPx,
+  @{lib}/geoclue-2.0/demos/agent      rPx,
+  @{bin}/python3.@{int}               rPx,
+  @{lib}/python3.@{int}               rPx,
```
roddhjav, line 51: - There is no profile for `@{bin}/python3.@{int}`
- There is no such thing as an exec name: `@{lib}/python3.@{int}`

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+
+  /etc/xdg/                           r,
+  /etc/xdg/autostart/                 r,
+  /etc/xdg/autostart/*.desktop        r,
+  /etc/xdg/menus/lxqt-*               r,
+  /etc/xdg/openbox/*                  r,
+  /etc/udev/udev.conf                 r,
+
+  owner @{HOME}/.local/share/         r,
+  owner @{HOME}/.config/              r,
```
roddhjav, line 77: Remove, they are included in `lxqt`

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+  owner @{HOME}/.config/autostart/    r,
+  owner @{HOME}/.config/autostart/*   rw,
+  owner @{user_cache_dirs}/openbox/   rw,
+  owner @{user_cache_dirs}/openbox/sessions/          rw,
+  owner @{user_cache_dirs}/openbox/openbox.log        rwk,
+  owner @{user_config_dirs}/mimeapps.list{,.@{rand6}} rw,
+  owner @{user_config_dirs}/dconf/user r,
+  owner @{user_config_dirs}/openbox/rc.xml            r,
+  owner @{user_share_dirs}/sddm/xorg-session.log      rw,
```
roddhjav, line 86: Remove, this is already part of `X-strict`

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+
+  @{PROC}/                             r,
+  @{PROC}/uptime                       r,
+  @{PROC}/@{pid}/stat                  r,
+  owner @{PROC}/@{pid}/stat            r,
+
+  @{run}/systemd/inhibit/**            rw,
```
roddhjav, line 93: That should be `@{att}/@{run}/systemd/inhibit/@{int}.ref rw,` and the rule should come before the `@{PROC}` ones.

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+  /dev/tty                             rw,
+
+  include if exists <local/lxqt-session>
+
+  profile systemctl {
+    include <abstractions/base>
+    include <abstractions/app/systemctl>
+
+    include if exists <local/lxqt-session_systemctl>
+
```
roddhjav, line 104: You forgot to close the systemctl subprofile: `}`

apparmor.d/groups/lxqt/lxqt-session
```diff
@@ -0,0 +1,115 @@
...
+  @{PROC}/                             r,
+  @{PROC}/uptime                       r,
+  @{PROC}/@{pid}/stat                  r,
+  owner @{PROC}/@{pid}/stat            r,
+
+  @{run}/systemd/inhibit/**            rw,
+
+  /dev/tty                             rw,
+
+  include if exists <local/lxqt-session>
```
roddhjav, line 97: Please more this include after the subprofile (it is where we place it for all other profiles).

roddhjav:
That should be good as a first version of the profile. Note that this part of lxqt is most likelly an app launcher. If so, you can remove all `rPx` rule and add `include <abstractions/app-launcher-user>`

apparmor.d/groups/lxqt/lxqt-session
Besanon: for now i will remove this, yes. I used a different profile for python. Only applicable with some backup programs

apparmor.d/groups/lxqt/lxqt-session
Besanon: didnt reflected this change, going to check the other profiles as well

Besanon:
which profile is supposed to be an app-launcher has been a bit confusing in the past. I will try to change this in lxqt-session, see if its ok to entirely remove the singel rPx-calls

Besanon:
well, i added   
include <abstractions/app-launcher-user> 
it is now in lxqt-session and lxqt-panel. I am not sure about which programs still need to be called via rPx. There are some. Maybe not all are reflected in the changes. But it is working. Have a look

PR #591, opened 2024-10-27
Add support for Xray and V2ray

EricLin0509:


apparmor.d/profiles-s-z/v2ray
```diff
@@ -0,0 +1,32 @@
...
+  network inet raw,
+  network inet6 dgram,
+  network inet6 raw,
+  network netlink raw,
+
+  @{exec_path} mr,
+
+  /etc/v2ray/ r,
+  /etc/v2ray/**.json r,
+  /usr/share/v2ray/**.dat r,
```
roddhjav, line 25: Simplify this as `/usr/share/v2ray/` it will be easier to maintain it.

EricLin0509: OK, fixed.

roddhjav:
Thanks, merged!

PR #603, opened 2024-11-09
Add unbound

barmogund:


apparmor.d/profiles-s-z/unbound
```diff
@@ -0,0 +1,64 @@
...
+
+@{exec_path} = @{bin}/unbound
+profile unbound @{exec_path} flags=(attach_disconnected) {
+  include <abstractions/base>
+  include <abstractions/bus-system>
+  include <abstractions/bus/org.freedesktop.login1>
+  include <abstractions/nameservice-strict>
+  include <abstractions/p11-kit>
+  include <abstractions/ssl_certs>
+  include <abstractions/common/systemd>
```
roddhjav, line 18: This abstraction is reserved for the systemd suite.

apparmor.d/profiles-s-z/unbound
```diff
@@ -0,0 +1,64 @@
...
+
+  capability net_bind_service,
+  capability net_raw,
+  capability sys_resource,
+  capability sys_chroot,
+  capability setgid,
+  capability setuid,
+  capability setpcap,
```
roddhjav, line 26: According to https://github.com/NLnetLabs/unbound/blob/733d5f7161c5675e66e39974b912a08eeb48140c/contrib/unbound.service.in#L56

setpcap is not needed.

apparmor.d/profiles-s-z/unbound
```diff
@@ -0,0 +1,64 @@
...
+  network unix stream,
+  network unix dgram,
+  network inet dgram,
+  network inet stream,
+  network inet6 dgram,
+  network inet6 stream,
+  network netlink raw,
+
+  signal (receive) set=(term, cont) peer=runsv,
+  #aa:dbus own bus=system name=org.freedesktop.resolve1
```
roddhjav, line 37: That cannot be `own`. You probably meant `talk`.

apparmor.d/profiles-s-z/unbound
```diff
@@ -0,0 +1,64 @@
...
+
+  @{exec_path} mr,
+
+  /etc/unbound/{,**} rw,
+  /etc/systemd/resolved.conf r,
+  /etc/systemd/resolved.conf.d/{,*} r,
+
+  @{run}/systemd/netif/links/* r,
+  @{run}/systemd/notify rw,
+  @{run}/systemd/resolve/{,**} rw,
```
roddhjav, line 52: Only `systemd-resolved` needs this. Are you sure about this?

apparmor.d/profiles-s-z/unbound
```diff
@@ -0,0 +1,64 @@
...
+  #aa:dbus own bus=system name=org.freedesktop.resolve1
+
+  dbus send bus=system path=/org/freedesktop/DBus
+       interface=org.freedesktop.DBus
+       member={GetConnectionUnixUser,GetConnectionUnixProcessID}
+       peer=(name=org.freedesktop.DBus, label=dbus-system),
+
+  @{exec_path} mr,
+
+  /etc/unbound/{,**} rw,
```
roddhjav, line 46: Not sure that `rw` is needed.

roddhjav:
As I mentioned in an in-line comment. This profile looks like a copy paste from `systemd-resolved`. Even if they are similar in concept. It is **not** how profile should be created.

barmogund:
well, yes, but see - not everyone may be able or just have the time to create every profile from scratch. Even if we construct them that way there are nearly always interferences with the whole which need to be considered afterwards.

roddhjav:
No, it is not how profiles should be created. Your original profile is a proof of it. Eg: by looking at the unbound source code, dbus is not needed.

PR #607, opened 2024-11-18
Add profile linter

nobody43:
Abstractions/tunables checks weren't done, they will take a while.
I don't know how you want to integrate it, so leaving it up to you. As well as applying the suggestions (or tell me how you see it being handled).
I will work in the same direction, after 3 weeks.

Related roddhjav#570

### DIFF
--- a/tests/profile_check.py
+++ b/tests/profile_check.py
@@ -456,6 +456,9 @@ def main(argv):
         profile_data.update(profilesInFile)
 
     for m in messages:
+        if m.get('suggestion'):
+            if m['suggestion'].endswith('\n'):
+                m['suggestion'] = m.get('suggestion').removesuffix('\n')
         m['filename'] = str(m.get('filename'))
         print(json.dumps(m, indent=2))
 
--- a/tests/profile_check.py
+++ b/tests/profile_check.py
@@ -186,14 +186,8 @@ def readApparmorFile(fullpath):
                     if m.get('profile'):
                         nestingStacker.append(m.get('profile'))  # set early
 
-                    if m.get('attachment') != '@{exec_path}' and not gotAttach:  # can be only singular
+                    if m.get('attachment') == '@{exec_path}' and not gotAttach:  # can be only singular
                         gotAttach = True
-                        messages.append({'filename':   fullpath,
-                                         'profile':    getCurrentProfile(nestingStacker),
-                                         'severity':   'WARNING',
-                                         'line':       n,
-                                         'reason':     "'@{exec_path}' must be defined as main path attachment",
-                                         'suggestion': None})
                         
                     profileMsg = {'filename':   fullpath,
                                   'profile':    getCurrentProfile(nestingStacker),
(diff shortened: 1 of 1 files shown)


tests/profile_check.py
```diff
@@ -0,0 +1,463 @@
...
+                'suggestion': suggestion_})
+    else:
+        msg = None
+
+    return msg
+
+def readApparmorFile(fullpath):
+    '''AA file could contain multiple AA profiles'''
+    headers = (
+        '# AppArmor.d - Full set of apparmor profiles',
```
nobody43, line 101: Modernize the name? (`apparmor.d` -> `AppArmor.d`)

roddhjav:
Nice work thanks a lot. It is really useful to have you working on this.

I think that for now, the your current structure will do it. Ideally (later), it would be nice to provide a proper `aa-lint` tool. I think we should try to make a linter that is not specific for this project (and that can ignore some check).

I think the implementation of `aa-lint` should look like existing linter in python such as [yamllint](https://github.com/adrienverge/yamllint). Thus, you will have to create a new directory in the apparmor.d project, e.g.: `aa-lint` with the python code.

I am going to create a new issue where we can discuss both the structure of the linter and the list of checks it should implement.

tests/profile_check.py
nobody43: OK, reverted.

tests/profile_check.py
```diff
@@ -0,0 +1,465 @@
...
+        from apparmor.rule.file     import FileRule, FileRuleset
+        from apparmor.common        import convert_regexp
+        try:
+            from apparmor.rule.variable import separate_vars
+        except ModuleNotFoundError:
+            from apparmor.aa            import separate_vars
+ 
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError(f"""Can't find 'python3-apparmor' package! Install with:
+$ sudo apt install python3-apparmor""")
```
roddhjav, line 463: This should be move outside the main with other import.

See, for example: https://github.com/roddhjav/pass-import/blob/19805430712d3b8b82f58373e3771d21bbe1fa64/pass_import/formats/kdbx.py#L10-L16

Then in the main: https://github.com/roddhjav/pass-import/blob/19805430712d3b8b82f58373e3771d21bbe1fa64/pass_import/formats/kdbx.py#L216

nobody43:
>That does not sound normal...

That was because of trailing '\n'. What would be the policy - add to suggestion or remove from the line?

>Suggesting owner for some file pattern is a good thing, however, it can raise some issues.

Should I remove it?

roddhjav:
We don't care if the vim syntax has a trailing line, so remove it. To be more precise, this is:
1. Not a bit issue, so let's not annoy profile developer with that
2. Not what has been done in the profile and thus, it would require to update all of them.

nobody43:
> We don't care if the vim syntax has a trailing line, so remove it.

I don't think that's possible while preserving the check. Each line in file have trailing `\n` at the end. While making a comparison I have to preserve it. Similar issue when taking original line as suggestion - see with output of `handleFileMessages()`.

> It is a full path, I manually anonymized the data.

Then you should delete the comment entirely because of edit history.

tests/profile_check.py
```diff
@@ -0,0 +1,480 @@
...
+                         'suggestion': abi})
+
+    # Ensure singular '@{exec_path}'
+    if not gotAttach:
+        messages.append({'filename':   fullpath,
+                         'profile':    None,
+                         'severity':   'WARNING',
+                         'line':       None,
+                         'reason':     "'@{exec_path}' must be defined as main path attachment",
+                         'suggestion': None})
```
nobody43, line 328: Should we ignore orphan profiles like `apparmor.d/groups/children/`? (in future PRs)

roddhjav: Yes, they will have to be ignored.

nobody43: `profiles-m-r/pam_roles` move to `profiles-m-r/pam/` and also exclude?

roddhjav:
Thanks, merged!

For future PR, try to follow the same architecture as [yamllint](https://github.com/adrienverge/yamllint).

PR #612, opened 2024-11-22
add profile for swayimg

valoq:


roddhjav:
It needs to be set in the attachment to: `profile swayimg @{exec_path} {`

roddhjav:
Good, thanks, merged.

PR #629, opened 2024-12-17
non-owner accesses authorized_keys

beroal:
When I try to log into a local SSH server, I get the following log messages:

```
apparmor="DENIED" operation="open" class="file" profile="sshd" name="/home/remote/.ssh/authorized_keys"  comm="sshd-session" requested_mask="r" denied_mask="r" fsuid=1005 ouid=0 FSUID="remote" OUID="root"
apparmor="DENIED" operation="exec" class="file" profile="sshd" name="/usr/bin/userdbctl"  comm="sshd-session" requested_mask="x" denied_mask="x" fsuid=0 ouid=0 FSUID="root" OUID="root"
```

After removing `owner` in

```
  owner @{HOME}/@{XDG_SSH_DIR}/authorized_keys{,.*} r,
```

no log messages.

### DIFF
--- a/apparmor.d/groups/ssh/sshd
+++ b/apparmor.d/groups/ssh/sshd
@@ -94,7 +94,7 @@ profile sshd @{exec_path} flags=(attach_disconnected) {
   owner @{user_download_dirs}/{,**} rwl,
   owner @{user_sync_dirs}/{,**} rwl,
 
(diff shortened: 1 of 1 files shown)


roddhjav:
Thanks, it should be fixed now.

Note: the sshd profile will soon have a major update to better handle the new sshd-session binary. Will have to double check this issue is not raised again.

PR #639, opened 2025-01-03
Update profile for xray

EricLin0509:
apparmor="DENIED" operation="open" class="file" profile="xray" name="/usr/share/v2ray/geosite.dat"  comm="xray" requested_mask="r" denied_mask="r" fsuid=956 ouid=0 FSUID="xray" OUID="root"


roddhjav:
Thanks, merged!

PR #645, opened 2025-01-14
Various fixes: man, nvidia-strict, iwd, lynx

el-remph:
* man:
  * more flexible index db path and specific manpath
  * allow calling groff and friends with inherited permissions
  * proper subprofiles
* nvidia-strict: nvidia libraries get argv from /proc/self
* iwd:
  * sends lots of UDP on packet sockets
  * writes to accept_ra and optimistic_dad
  * DNS daemons other than systemd-resolvd may require iwd to use resolvconf
* lynx: now handles brotli'd HTML as well as gzipped

I didn't want to open a whole pull request for each profile given that each change is just a few lines, but let me know if I should

### DIFF
--- a/apparmor.d/profiles-g-l/lynx
+++ b/apparmor.d/profiles-g-l/lynx
@@ -33,8 +33,7 @@ profile lynx @{exec_path} {
   /etc/mailcap r,
   /etc/mime.types r,
 
-  owner @{tmp}/lynxXXXX*/ rw,
-  owner @{tmp}/lynxXXXX*/*TMP.html{,.gz} rw,
+  owner @{tmp}/lynxXXXX*/{,**} rw,
 
   include if exists <local/lynx>
 }
--- a/apparmor.d/abstractions/nvidia-strict
+++ b/apparmor.d/abstractions/nvidia-strict
@@ -26,7 +26,8 @@
         @{PROC}/sys/vm/max_map_count r,
         @{PROC}/sys/vm/mmap_min_addr r,
         @{PROC}/modules r,
-  owner @{PROC}/@{pid}/comm r,
+  owner @{PROC}/@{pid}/{,task/*/}comm r,
+  owner @{PROC}/@{pid}/cmdline r,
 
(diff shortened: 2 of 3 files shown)


apparmor.d/profiles-g-l/lynx
```diff
@@ -34,7 +34,8 @@ profile lynx @{exec_path} {
   /etc/mime.types r,
 
   owner @{tmp}/lynxXXXX*/ rw,
-  owner @{tmp}/lynxXXXX*/*TMP.html{,.gz} rw,
+  # soon there may be .zst here as well. Maybe just give it the whole dir?
+  owner @{tmp}/lynxXXXX*/*TMP.html{,.{gz,br}} rw,
```
roddhjav, line 38: That should be simplified as:
```
owner @{tmp}/lynxXXXX*/{,**} rw,
```

roddhjav:
Thanks for you PR.

Regarding the nvidia-strict abstraction, I add rules to this abstraction with extra care. Can you tell me more why you think they are needed for all Nvidia GPU access?

Regarding `man`: This profile is disabled (the actual profile in use is the one shipped by the default with the man package). I won't accept update as it is not the one caused the issue.

el-remph:
Rebased without man fixes and with lynx simplification.

>Regarding the nvidia-strict abstraction, I add rules to this abstraction with extra care. Can you tell me more why you think they are needed for all Nvidia GPU access?

- /proc/self/task/comm allows nvidia library to see /proc/self/comm (which it is already allowed) if it has been changed by the calling process for an individual thread.
- /proc/self/cmdline allows an nvidia library to see the argv of the calling process. Since an nvidia library is in the same process space as its caller there is nothing to stop it just reading argv directly, so I guess allowing /proc/self/cmdline is just more convenient, disallowing it would be no more secure.

>Regarding man: This profile is disabled (the actual profile in use is the one shipped by the default with the man package). I won't accept update as it is not the one caused the issue.

I'll go to upstream then; which package did you mean, the debian package?

PR #659, opened 2025-01-26
Update pacman

curiosityseeker:
profile pacman//systemctl {
  signal send set=(cont term) peer=systemd-tty-ask-password-agent,
}


roddhjav:
Thanks, merged!

PR #663, opened 2025-01-26
Update dkms

curiosityseeker:


apparmor.d/profiles-a-f/dkms
```diff
@@ -47,10 +50,14 @@ profile dkms @{exec_path} flags=(attach_disconnected) {
   @{lib}/linux-kbuild-*/scripts/**             rix,
   @{lib}/linux-kbuild-*/tools/objtool/objtool  rix,
   @{lib}/llvm-[0-9]*/bin/clang                 rix,
+  @{lib}/modules/*/build/arch/x86/**           rix,
```
roddhjav, line 53: Use the `@{arch}` variable here.

curiosityseeker: I had considered this. However, `@{arch}` is defined as:

`@{arch}=x86_64 amd64 i386 i686`

and that doesn't fit.

roddhjav:
Thanks, merged!

PR #669, opened 2025-02-22
Fix broken regex in mkinitcpio profile

c-jaenicke:
This PR adds one extra star in the [mkinitcpio profile on line 65](https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/groups/pacman/mkinitcpio#L65), fixing the broken regex.

The broken regex causes the AppArmor service to fail to start and the mkinitcpio profile to be unavailable.  Detected while doing a fresh install on Arch Linux on the 22.02.2025.


roddhjav:
Thanks, merged!

PR #707, opened 2025-03-26
read access to the hwmon directory

beroal:
Without this patch, I get
```
sensors_init: Kernel interface error
```


roddhjav:
Thanks merged!

Issue #716, opened 2025-04-10
Add arm architecture to PKGBUILD


zinootje:
Add the moment only the x86_64 arch is specified. Could this be changed to any to also support arm and others. To my understanding i don't think the profiles are architecture dependent.

PR #717, opened 2025-04-10
Update PKGBUILD arch to any

zinootje:
updated PKGBUILD arch to any to support all archs
see issue #716


roddhjav:
As mentioned in #716. You need to add an arm arch such as `aarch64` instead of setting `any`.

roddhjav:
Thanks, merged!

PR #729, opened 2025-04-25
webcam

beroal:


beroal:
> That should be `@{run}/udev/data/c81:@{int} r, # For video4linux` instead. See https://raw.githubusercontent.com/torvalds/linux/master/Documentation/admin-guide/devices.txt (search for "81 char")

Done.

PR #746, opened 2025-05-18
Fix hyprland profile.

tpaau-17DB:
Hyprland now needs access to `~/.local/share/hyprland/`, where it stores its last version used for notifying user when it was updated. Without `rw` permissions to this directory, Hyprland displays a "Hyprland was just updated!" notification on every startup. This PR resolves that issue.


roddhjav:
Thanks, merged!

PR #761, opened 2025-06-11
Add xkeyboard-config-2 ressources to desktop abstraction

### LINKED ISSUE #760: GDM and Gnome-shell couldn't start after updating xkeyboard-config
GDM and Gnome-shell couldn't start after updating [xkeyboard-config](https://archlinux.org/packages/extra/any/xkeyboard-config/)

Here are the Logs:

```
apparmor="DENIED" operation="capable" class="cap" profile="gdm"  comm="gdm" capability=21  capname="sys_admin"
apparmor="DENIED" operation="open" class="file" profile="gnome-shell" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="Mutter Input Th" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gnome-session-binary" name="/usr/share/gnome-shell/extensions/"  comm="gnome-session-b" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="xwayland" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="Xwayland" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-wacom" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-wacom" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-color" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-color" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-keyboard" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-keyboard" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-keyboard" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-keyboard" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-power" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-power" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-media-keys" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-media-keys" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="create" class="net" profile="gsd-media-keys"  comm="gsd-media-keys" family="inet" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-media-keys"  comm="gsd-media-keys" family="inet6" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="open" class="file" profile="gsd-wacom" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-wacom" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="open" class="file" profile="gsd-color" name="/usr/share/xkeyboard-config-2/rules/evdev"  comm="gsd-color" requested_mask="r" denied_mask="r" fsuid=120 ouid=0 FSUID="gdm" OUID="root"
apparmor="DENIED" operation="create" class="net" profile="gsd-keyboard"  comm="gsd-keyboard" family="netlink" sock_type="raw" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-media-keys"  comm="gsd-media-keys" family="inet6" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-keyboard"  comm="gsd-keyboard" family="inet" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-keyboard"  comm="gsd-keyboard" family="inet6" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-keyboard"  comm="gsd-keyboard" family="inet6" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-power"  comm="gsd-power" family="inet" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-power"  comm="gsd-power" family="inet6" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-power"  comm="gsd-power" family="inet6" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-color"  comm="gsd-color" family="netlink" sock_type="raw" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-color"  comm="gsd-color" family="inet" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-color"  comm="gsd-color" family="inet6" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-color"  comm="gsd-color" family="inet6" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-wacom"  comm="gsd-wacom" family="netlink" sock_type="raw" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-wacom"  comm="gsd-wacom" family="inet" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-wacom"  comm="gsd-wacom" family="inet6" sock_type="dgram" protocol=0 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-wacom"  comm="gsd-wacom" family="inet6" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="create" class="net" profile="gsd-wacom"  comm="gsd-wacom" family="inet" sock_type="stream" protocol=6 requested_mask="create" denied_mask="create"
apparmor="DENIED" operation="signal" class="signal" profile="gdm-session-worker"  comm="gdm-session-wor" requested_mask="send" denied_mask="send" signal=hup peer="unconfined"
```

valoq:
Fixes #760

### DIFF
--- a/apparmor.d/abstractions/desktop
+++ b/apparmor.d/abstractions/desktop
@@ -77,6 +77,7 @@
   /usr/share/desktop-base/{,**} r,
   /usr/share/hwdata/*.ids r,
   /usr/share/icu/@{int}.@{int}/*.dat r,
+  /usr/share/xkeyboard-config-2/{,**} r,
 
   include if exists <abstractions/desktop.d>
 


curiosityseeker:
Unfortunately, this fix is only a first step as it is not sufficient. In my case xwayland, xorg, kwin_wayland, systemd-localed were affected, and with the exception of xorg the desktop abstraction is not included in those profiles.

valoq:
> Unfortunately, this fix is only a first step as it is not sufficient. In my case xwayland, xorg, kwin_wayland, systemd-localed were affected, and with the exception of xorg the desktop abstraction is not included in those profiles.

If I  am not overlooking something, they all include either the wayland or X-strict abstraction, which are also both in the desktop abstraction. In case of systemd-localed the profile includes a lot of X11 resources defined there that are already in other X11 related abstractions, which probably means the profile should get an overhaul itself.

We could add the new permissions to both the wayland and X-strict abstraction, but I am wondering if it would be better to simply include desktop in all profiles that currently include those two abstractions

roddhjav:
Thanks for this PR, I can confirm the issue.

Regarding the fix, `/usr/share/xkeyboard-config-2/` is literally the new path for `/usr/share/X11/xkb` (`/usr/share/X11/xkb` is now a link to `/usr/share/xkeyboard-config-2/`). Therefore, the rule`/usr/share/xkeyboard-config-2/{,**} r,` should be added at the same place as the same location as `X11/xkb`, that is to said in:
- `abstractions/X-strict`
- `software-properties-gtk`
- `systemd-localed`
- `xkbcomp`

Can you update this PR accordingly

valoq:
> Thanks for this PR, I can confirm the issue.
> 
> Regarding the fix, `/usr/share/xkeyboard-config-2/` is literally the new path for `/usr/share/X11/xkb` (`/usr/share/X11/xkb` is now a link to `/usr/share/xkeyboard-config-2/`). Therefore, the rule`/usr/share/xkeyboard-config-2/{,**} r,` should be added at the same place as the same location as `X11/xkb`, that is to said in:
> 
>     * `abstractions/X-strict`
> 
>     * `software-properties-gtk`
> 
>     * `systemd-localed`
> 
>     * `xkbcomp`
> 
> 
> Can you update this PR accordingly

Done

Note that xkbcomp already includes the abstraction X-strict and therefore does not need the permission.

valoq:
I found that imv is also broken unless the new permissions are also added to the desktop abstraction

roddhjav:
That is not needed: the desktop abstraction include `X-strict`, therefore the rule is already included.

roddhjav:
Thanks; merged!

valoq:
> That is not needed: the desktop abstraction include `X-strict`, therefore the rule is already included.

Right, fixed that again.

There is however a general issue regarding the related X-strict file.
When running applications under wayland with X-wayland disabled, many applications still need access to X resources like` /usr/share/X11/` and `/usr/share/xkeyboard-config-2/` imv-wayland being one such example.
As long as X-strict is in the desktop desktop abstraction that is not a problem, but for running a pure wayland desktop, the X-strict abstraction should be removable, which means that  the access permissiosn for the X11 ressources needed by wayland applications should be available in a place like the desktop abstraction

roddhjav:
I agree with you on this matter; I actually think it is why xkeyboard has been moved outside `/usr/share/X11/`, because it is also required for wayland. This will be fully revisited when support for condition get into apparmor profile.

PR #770, opened 2025-06-18
Stacking firefox-crashhelper

curiosityseeker:
DENIED  firefox exec @{lib}/firefox/crashhelper -> firefox-crashhelper info="no new privs" comm=firefox requested_mask=x denied_mask=x error=-1

### DIFF
--- a/apparmor.d/abstractions/app/firefox
+++ b/apparmor.d/abstractions/app/firefox
@@ -58,7 +58,6 @@
 
   @{lib_dirs}/{,**}             r,
   @{lib_dirs}/*.so              mr,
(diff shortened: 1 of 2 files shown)


roddhjav:
This need to be moved to the firefox profile, and stacked in this profile. Otherwise, it won't work well with other profile using the firefox abstraction as base (thunderbird and the torbrowser).

curiosityseeker:
Yes, you are right. So I changed that accordingly. Note, that I didn't test, e.g., the `torbrowser` profile if it is affected of the removal in the abstraction as I don't use that browser.

roddhjav:
Thanks, merged!

Regarding the torbrowser, it does not matter: these things are disabled in it.

PR #778, opened 2025-06-29
Ouch: Allow listing archives

valoq:
The list function requires the creation of new directories in /tmp for some archive formats


valoq:
please delete (messed up branch)

PR #779, opened 2025-06-29
ouch: allow listing archive contents

valoq:
The list function requires the creation of new directories in /tmp for some archive formats


roddhjav:
Thanks, merged!

PR #780, opened 2025-07-01
Add profile for `tgpt`.

tpaau:
## New profile for [tgpt](https://github.com/aandrew-me/tgpt)

The profile is set to unconfined at this time since it's very experimental and *will* break some functionality.

While set to enforce mode, with the unconfined flag removed:
- [x] Basic LLM chat works
- [x] Image generation works, as long as the output path is in `XDG_DOWNLOAD_DIR`
- [ ] Shell commands can be generated, but can't be executed as allowing the program to run a shell in unconfined mode would be to defeat the purpose of having an AppArmor profile for it in the first place

### DIFF
--- a/apparmor.d/profiles-s-z/tgpt
+++ b/apparmor.d/profiles-s-z/tgpt
@@ -7,7 +7,7 @@ abi <abi/4.0>,
 include <tunables/global>
 
 @{exec_path} = @{bin}/tgpt
-profile tgpt @{exec_path} flags=(unconfined) {
+profile tgpt @{exec_path} flags=(complain) {
   include <abstractions/base>
   include <abstractions/nameservice>
 


roddhjav:
I am not going to add unconfined profile to this repo. You have to use the complain flag for dev.

Now this program is a bit tricky, there is a trade-off between security and compatibility to make; depending on this, some part of the the shell could be confined to prevent some access (eg su, sudo...).

tpaau:
I don't think there is a secure way of letting the program to run shell code. 

Yes, you could restrict its access to `suid` and other critical files, but then users will still ask questions like "How do I update my system?" or "How do I install an app?". I'd simply block shell commands.

It's really not that big of a deal to do a Ctrl+Shift+C and a Ctrl+Shift+V to execute the command.

PR #785, opened 2025-07-16
Add profile for texstudio

valoq:
Add a profile for the latex editor texstudio

The applications tries to open some file under /usr/share/ with write permissions for some reason.
I assume this is a bug since it will never have those permissions anyway, therefore I silenced the request in the profile.


roddhjav:
Can you fix the CI issues, otherwise, it LGTM.

roddhjav:
Thanks, merged!

PR #791, opened 2025-07-23
Add profiles for wget and curl

valoq:
My initial tests show no more complains in the audit logs after going through common example uses, but some additional testing may be advisable.

### DIFF
--- a/apparmor.d/profiles-s-z/curl
+++ b/apparmor.d/profiles-s-z/curl
@@ -1,0 +1,31 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2025 valoq <valoq@mailbox.org>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/curl
+profile curl @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/ssl_certs>
+  include <abstractions/p11-kit>
+  include <abstractions/nameservice-strict>
+  include <abstractions/user-download-strict>
+
+  network inet dgram,
+  network inet6 dgram,
+  network inet stream,
+  network inet6 stream,
+  network netlink raw,
+
+  @{exec_path} mr,
+
+  owner @{HOME}/.curlrc r,
+  owner @{user_config_dirs}/curlrc r,
+
+  include if exists <local/curl>
+}
+
+# vim:syntax=apparmor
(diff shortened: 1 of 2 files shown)


roddhjav:
Thanks, for your PR. However, I won't add these profiles as part of this project.

Here is why:
1. These profiles are special one: to prevent breaking someone computer, you must allow downloading file everywhere. Therefore, you need `/** rw,`. And this has consequences for the profile's security (obviously) but also for other's (if you allow transition to curl/wget). Take a look at this PR on apparmor where I explain the differences between weak/strong confinement: https://gitlab.com/apparmor/apparmor/-/merge_requests/1485#note_2306531478
1. There is an open PR upstream for wget, (with more tests). See https://gitlab.com/apparmor/apparmor/-/merge_requests/1699

valoq:
Not sure I follow your reasoning here;
Applications like wget and curl can be used in many places and therefore could be argued to required file access to everything?
But the same is true for most applications like say image viewers trying to read a file in ${HOME}/mydir or /etc/sway/mywallpaper.png
If I recall correctly, your argumentation in a different discussion was that this project follows the Android model and expects certain files in certain locations (e.g. images in ~/Pictures)
Wouldn't the same be true for a download program that would have to be run in ~/Downloads then?

Or is this about the interactive use within scripts? That make more sense I suppose, though I wonder if this can be avoided, e.g. by restricting the application only in interactive use?

roddhjav:
Yes, a profile even *weak* is better than no profile at all. See https://gitlab.com/apparmor/apparmor/-/merge_requests/1485#note_2312844267

> But the same is true for most applications like say image viewers trying to read a file in ${HOME}/mydir or /etc/sway/mywallpaper.png

It is different because curl/wget is not only a user application that can legitimately only access labelled files. They can be used by a lot of progam/scripts that are not confined in this project. In a default mode (not FSP, RBAC), we cannot provide profiles that would break this kind of tool.

To ensure both compatibility and security (ie: still providing some protection), we need to put a big [`/** rw`](https://gitlab.com/apparmor/apparmor/-/merge_requests/1699/diffs#e81828620af30199ddc07221efb6614576601c6e_0_49) in the profile. And it [does change everything](https://gitlab.com/apparmor/apparmor/-/merge_requests/1485#note_2306531478): if you allow transition from a normal profile to this one, you allow your normal profile full write access to the system. Thus, with only 1 profile of this kind, if you do not take extra care of it, you can literally destroy the full security of your project.

I am currently working with upstream on this matter, I do not want to merge this kind of profile here before we agree on how we should handle them.

Later, they will get merged (because handled correctly they do provide security)

valoq:
I'm wondering if it would make sense to introduce profiles that are meant to run only in complain mode. It would still help with tracing access in the audit logs even if some profiles cannot be effectively be enforced. wget and curl being some examples

PR #794, opened 2025-07-27
add xournalpp

valoq:
Adds a profile for xournalpp

The applications includes audio recording features which appears to require direct write access to some device files in /dev/snd


roddhjav:
Thanks, I have tested it, it looks good.

A few minor comments:
- Move the base abstraction first.
- No need of `abstractions/fonts` as you have `abstractions/desktop`
- Allow xournalpp to create its own config & cache. E.g. `owner @{user_config_dirs}/xournalpp/{,**} rw,`
- Move `/dev/snd/` after the PROC rule.
- You need to add: `  @{open_path} rPx -> child-open-browsers,` to allow xournalpp to open its manual page.

roddhjav:
Good, merged.

PR #801, opened 2025-08-06
add profiles for wayland screen capture tools

valoq:
Add profiles for wayland screen capture tools grim and slurp


roddhjav:
Thanks, sorry for the delay, they looks good to me merged.

PR #805, opened 2025-08-07
Add Pacman Hook Profile for Paccache

RayJW:
This adds a profile for the [paccache-hook](https://aur.archlinux.org/packages/paccache-hook) and the tool `paccache` it's using, which is part of [pacman-contrib](https://archlinux.org/packages/extra/x86_64/pacman-contrib).

The initial profile created by `aa-log -r pacman` contained a lot of specific paths to packages and specific pacman config files. I think I've cleaned all of that up. It seems like a pretty big profile, but testing went fine on my end. This is my first time creating a profile from scratch though, so maybe this is way off, but I tried following the guide.

### DIFF
--- a/apparmor.d/groups/pacman/pacman-hook-paccache
+++ b/apparmor.d/groups/pacman/pacman-hook-paccache
@@ -1,5 +1,5 @@
 # apparmor.d - Full set of apparmor profiles
-# Copyright (C) 2025 valoq <valoq@mailbox.org>
+# Copyright (C) 2025 RayJW <29835902+RayJW@users.noreply.github.com>
 # SPDX-License-Identifier: GPL-2.0-only
 
 abi <abi/4.0>,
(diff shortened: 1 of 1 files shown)


roddhjav:
It is not how profiles works. Beware, aa-log only converts logs to rule, they can't be copied and pasted like this.

Have a look at the existing apparmor profiles to see an example. For instance, as you want to confine a pacman hook: [pacman-hook-dkms](https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/groups/pacman/pacman-hook-dkms). So you can create the base profile for your hook, named `pacman-hook-paccache`:
```
@{exec_path} = /usr/share/libalpm/scripts/paccache-hook.sh
profile pacman-hook-paccache @{exec_path} flags=(attach_disconnected) {
  include <abstractions/base>

  @{exec_path} mr,

  include if exists <local/pacman-hook-paccache>
}
```

Then, you can use `aa-log` to see the missing rules. There is already a profile for `paccache`, thus you can use it here. Most likely it will have to be used as `@{bin}/paccache rPx,` in `pacman-hook-paccache`.

Also, you may want to have a look at the doc: https://apparmor.pujol.io/development/workflow/

**Side note**: The rules you have are unrelated to `paccache-hook.sh` (if the profile doesn't exist, it runs unconfined). They mean that pacman wants to run paccache (probably from another hook or an install script). Add paccache in the pacman profile and it will be fixed.

PR #807, opened 2025-08-08
Host: allow netlink raw

doublez13:
Querying a DNS server using it's hostname results in an apparmor denial: `host google.com dns.google.com`

`apparmor="DENIED" operation="create" class="net" profile="host" pid=00000 comm="host" family="netlink" sock_type="raw" protocol=0 requested_mask="create" denied_mask="create"`


roddhjav:
Good catch thanks, merged

PR #810, opened 2025-08-11
Add basic profiles for kerberos client utilities

doublez13:
These profiles implement AppArmor policies for basic kerberos client functionality. 

There is certainly missing functionality (sssd-kcm integration for example), but I believe these should be sufficient for most use-cases.


apparmor.d/profiles-g-l/kdestroy
```diff
@@ -0,0 +1,29 @@
...
+include <tunables/global>
+
+@{exec_path} = @{bin}/kdestroy
+profile kdestroy @{exec_path} {
+  include <abstractions/base>
+
+  @{exec_path} mr,
+
+  #Allow root to destroy other users' creds cache
+  capability dac_override,
```
roddhjav, line 15: This needs to be moved before `@{exec_path} mr,` (same for other profiles)

apparmor.d/profiles-g-l/kdestroy
```diff
@@ -0,0 +1,29 @@
...
+
+  #Config Files
+  /etc/krb5.conf r,
+  /etc/krb5.conf.d/{,**} r,
+
+  #Credentials cache
+  /tmp/krb5cc_* rwk,
+  /tmp/tkt* rwk,
```
roddhjav: 1. Use the `kerberosclient` abstraction instead>
2. Can you precise the `*`, it may be `@{word6}` or similar.

apparmor.d/profiles-g-l/kinit
```diff
@@ -0,0 +1,39 @@
...
+
+  #Config Files
+  /etc/krb5.conf r,
+  /etc/krb5.conf.d/{,**} r,
+
+  #Host keytab file
+  /etc/krb5.keytab r,
+
+  #User keytab file
+  /var/lib/krb5/user/*/client.keytab r,
```
roddhjav: Can `*` be precised? `@{user}`, `@{uid}` ?

roddhjav:
Thanks, as they are still a WIP, can you set the complain in flag `main/dists/flags/main.flag`.

apparmor.d/profiles-g-l/kdestroy
doublez13: For point 2, this seems to differ between different distros.

On Debian and Arch, the formatting is `/tmp/krb5cc_@{uid}`
On RHEL Distros, the formatting seems to be `/tmp/krb5cc_@{uid}_@{word10}`

I'm hesitant to be too specific. Even the kerberosclient abstraction just specifies `/tmp/krb5cc*`

roddhjav:
Nice, thanks. merged!

PR #825, opened 2025-08-27
Add cider profile

Stoppedpuma:
Adds a profile for [Cider](https://cider.sh/). This is for the new *proprietary* version since development has ended on the old version that was open source a few years back. As this is proprietary, I would greatly appreciate a zero trust approach to this and point out any issues that would usually be negligible if there is anything that should be changed.


apparmor.d/profiles-a-f/cider
```diff
@@ -0,0 +1,61 @@
...
+  @{lib_dirs}/libffmpeg.so mr,
+  @{lib_dirs}/chrome-sandbox rpx,
+
+  @{bin}/xdg-settings rpx,
+
+  owner @{user_config_dirs}/sh.cider.genten/ rw,
+  owner @{user_config_dirs}/sh.cider.genten/** rwk,
+  owner @{user_config_dirs}/sh.cider.genten/WidevineCdm/*/_platform_specific/linux_x64/libwidevinecdm.so mr,
+  owner @{user_config_dirs}/sh.cider.genten/WidevineCdm/*/manifest.json r,
+  owner @{user_config_dirs}/sh.cider.genten/WidevineCdm/*/latest-component-updated-widevine-cdm r,
```
roddhjav: These rules are already included by the previous one.

apparmor.d/profiles-a-f/cider
```diff
@@ -0,0 +1,61 @@
...
+  @{lib_dirs}/ r,
+  @{lib_dirs}/** r,
+  @{lib_dirs}/libffmpeg.so mr,
+  @{lib_dirs}/chrome-sandbox rpx,
+
+  @{bin}/xdg-settings rpx,
+
+  owner @{user_config_dirs}/sh.cider.genten/ rw,
+  owner @{user_config_dirs}/sh.cider.genten/** rwk,
+  owner @{user_config_dirs}/sh.cider.genten/WidevineCdm/*/_platform_specific/linux_x64/libwidevinecdm.so mr,
```
roddhjav: `linux_x64` -> `linux_@{arch}`

apparmor.d/profiles-a-f/cider
```diff
@@ -0,0 +1,61 @@
...
+  network inet stream,
+  network inet6 stream,
+  network netlink raw,
+
+  @{exec_path} mrix,
+
+  @{lib_dirs}/ r,
+  @{lib_dirs}/** r,
+  @{lib_dirs}/libffmpeg.so mr,
+  @{lib_dirs}/chrome-sandbox rpx,
```
roddhjav: If you do not have a profile for this one, it won't work

apparmor.d/profiles-a-f/cider
```diff
@@ -0,0 +1,61 @@
...
+
+  @{bin}/xdg-settings rpx,
```
roddhjav: `rPx` instead of `rpx`.

apparmor.d/profiles-a-f/cider
```diff
@@ -0,0 +1,61 @@
...
+  @{PROC}/@{pid}/task/ r,
+  @{PROC}/@{pid}/task/@{tid}/status r,
+  @{PROC}/sys/fs/inotify/max_user_watches r,
+  owner @{PROC}/@{pid}/cgroup r,
+  owner @{PROC}/@{pid}/cmdline r,
+  owner @{PROC}/@{pid}/fd/ r,
+  owner @{PROC}/@{pid}/oom_{,score_}adj rw,
+  owner @{PROC}/@{pid}/statm r,
+
+  /usr/share/xkeyboard-config-2/** r,
```
roddhjav: That should be part of the wayland/X system. I am afraid that only this will fail on a lot of setup.

roddhjav:
>  I would greatly appreciate a zero trust approach

Zero trust does not mean anything in this context; however, by construction all profiles are always in the stricter position possible.

I do not know if Cider is still an electron based app. If yes, you should use `common/electron` instead

Stoppedpuma:
I've adjusted the profile to your review, please ensure these changes are correct.

roddhjav:
Good, merged!

PR #830, opened 2025-08-29
rpcbind: update profile

yukiteruamano:
Hi! 

This is a simple update for rpcbind profile. With this diff the profile work out-of-the-box for complain and enforce mode.


roddhjav:
Merged thanks

PR #832, opened 2025-08-29
firewall-applet: update profile

yukiteruamano:
Add support for user config files for firewall applet and cgroup info for pid process


roddhjav:
Merged thanks

PR #841, opened 2025-09-04
fix slurp

Stoppedpuma:


roddhjav:
You should use the wayland abstraction instead.
You probably also need to use the icon abs.

Stoppedpuma:
I've updated this to match your comment, though I'm unsure why the icons abstraction would be needed for slurp?

roddhjav:
Good merged!

> though I'm unsure why the icons abstraction would be needed for slurp?

The initial profile has: `/usr/share/icons/{,**} r,` meaning other icons location are needed too, hence the abstraction.
Looking at the source code, it is required by slurp for the cursor.

PR #847, opened 2025-09-09
add poppler tools

valoq:
This adds the poppler command line tools for manipulating pdf files

pdfattach
pdfdetach
pdffonts
pdfimages
pdfinfo
pdfseparate
pdfsig
pdftocairo
pdftohtml
pdftoppm
pdftops
pdfunite

### DIFF
--- a/apparmor.d/profiles-m-r/pdftoppm
+++ b/apparmor.d/profiles-m-r/pdftoppm
@@ -6,8 +6,8 @@ abi <abi/4.0>,
 
 include <tunables/global>
 
-@{exec_path} = @{bin}/pdftotext
-profile pdftotext @{exec_path} {
+@{exec_path} = @{bin}/pdftoppm
+profile pdftoppm @{exec_path} {
   include <abstractions/base>
   include <abstractions/user-read-strict>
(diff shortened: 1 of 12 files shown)


apparmor.d/profiles-m-r/pdftoppm
```diff
@@ -0,0 +1,25 @@
...
+  include <abstractions/base>
+  include <abstractions/fonts>
+  include <abstractions/user-read-strict>
+  include <abstractions/user-write-strict>
+
+  @{exec_path} mr,
+
+  /usr/share/poppler/{,**} r,
+
+  owner /tmp/{,**} rw,
```
roddhjav: The user-tmp abstraction is forbidden because allowing `owner /tmp/{,**} rw,` could allow confinement escape. Therefore; you need to precise what are the files needed.

valoq: Problem is that this tool is often used for previews and each applications uses an individual path to save them

I restricted the permissions to the supported image formats now and removed read access.

roddhjav:
Thans, merged

PR #849, opened 2025-09-10
add preview tools

valoq:
This adds multiple media/preview tools commonly used by tui file managers

- chafa
- ~~magick~~ (added by other MR)
- resvg
- ueberzugpp


curiosityseeker:
These are useful profiles, IMO. However, I wonder if they should be even more strict.

I'm only familiar with `magick` (and not with the other tools). It's used to "convert between image formats as well as resize an image, blur, crop, despeckle, dither, draw on, flip, join, re-sample, and much more" according to `man magick`.  So why not restrict it to these image formats by using the existing `@{image_ext}` variable and, in addition, `[pP][dD][fF]` as converting images to pdf is possible as well?

I know that this is problematic when using the `user-read-strict` and `user-write-strict` abstractions. So my idea would be to create a variable called, e.g., `@{user-strict}` somewhere in `tunables/multiarch.d` using the paths used in above abstractions.

Then we could add rules like:
```
owner @{user-strict}/**.[pP][dD][fF] rwkl,
owner @{user-strict}/**.@{image_ext} rwkl,
```
The first rule could also be used in your PR regarding the poppler tools. 

Or do you consider this overkill?

EDIT: Btw., shouldn't we use `base-strict` now by default?

roddhjav:
Using `**.@{image_ext}` could be a good option here. 

I agree that having something like `@{user_dirs}` could be an option too. But it will requires some tests and goes beyond the scope of this PR.

> EDIT: Btw., shouldn't we use base-strict now by default?

No, the `base-strict` abstraction is *already* used, there is a job that is converting the abs at build time.

valoq:
> Using `**.@{image_ext}` could be a good option here.

I would like to see something like this for several types of files, with A) something like the user-strict path as condition mentioned by @curiosityseeker and B) the general permission to access certain file types regardless of path and owner. Both should be entire abstractions like image-files-user and image-files-global

Other file types could be:

- document-files (read only) = pdf, ps, djvu, cb (generally read only document formats that are normally not edited)
- office-files (read-write) = odt, docx, xls, pdf (maybe include the document-files abstraction here)
- music-files (read only) = all audio formats
- video-files (read only) = all video formats
- music-files-edit (read-write) = all audio formats
- video-files-edit (read-write) = all video formats

That would be a different RP though.

curiosityseeker:
> I would like to see something like this for several types of files,

Well, there is already https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/tunables/multiarch.d/extensions. Together with a `@{user_dirs}` or `@{user_dirs_strict}` variable this would allow for very flexible rules, like:

```
/**.@{image_ext} r,
owner @{user_dirs_strict/**.@{image_ext} rwkl,
```

valoq:
I missed that one, thanks

Not sure if we want to use **.@{image_ext} here

valoq:
@curiosityseeker Do you want to create a PR for the @{user-strict}
For now I dont think its relevant to this PR but it would provide a significant enhancement for many profiles once available

curiosityseeker:
> @curiosityseeker Do you want to create a PR for the @{user-strict} 

Yes, I will.

roddhjav:
@curiosityseeker Ensure you use `@{user_dirs_strict}` (you cannot have `-` in variable, so `@{user-strict}` won't work).

roddhjav:
See https://github.com/roddhjav/apparmor.d/discussions/873

valoq:
Implemented by #906

The alternative MR uses the new file type variable instead of allowing access to all user files. See:
https://github.com/roddhjav/apparmor.d/pull/906/commits/7cec17331aa4b53c0ffd254a6181b703de071ebd

PR #850, opened 2025-09-10
add search tools

valoq:
Adds file (fd) and string (rg) search tools

This is a bit special:
These tools need access tot he entire file system for common usage, but since they parse a lot of data they should also be restricted. Since there are no permissions needed that could allow data extraction (like network access) this still provides significant benefits.

### DIFF
--- a/apparmor.d/profiles-a-f/fd
+++ b/apparmor.d/profiles-a-f/fd
@@ -1,0 +1,21 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2025 valoq <valoq@mailbox.ofd>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/fd
+profile fd @{exec_path}  {
+  include <abstractions/base>
+
+  @{exec_path} mr,
+
+  ## Allow reading the entire filesystem to search for filenames
+  /{,**} r,
+
+  include if exists <local/fd>
+}
+
+# vim:syntax=apparmor
--- a/apparmor.d/profiles-m-r/rg
+++ b/apparmor.d/profiles-m-r/rg
@@ -1,0 +1,21 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2025 valoq <valoq@mailbox.org>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/rg
+profile rg @{exec_path}  {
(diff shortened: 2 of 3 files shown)


roddhjav:
This is good.

They fit into the [recently defined](https://gitlab.com/apparmor/apparmor/-/merge_requests/1485#note_2306531478) concept of generic (vs narrow) profiles.

However, they need to be handled with care: allowing something like `@{bin}/rg Px,` from other profiles would allow a confinement escape. Thus, I need to find a way to handle them correctly (possible a new linter check)

apparmor.d/profiles-a-f/fd
```diff
@@ -0,0 +1,21 @@
...
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/fd
+profile fd @{exec_path}  {
+  include <abstractions/base>
+
+  ## Allow reading the entire filesystem to search for filenames
+  /{,**} r,
```
roddhjav, line 19: Move it after `@{exec_path}`

valoq: Done

curiosityseeker:
> However, they need to be handled with care: allowing something like `@{bin}/rg Px,` from other profiles would allow a confinement escape. 

I agree. So wouldn't adding the `deny-sensitive-home` and `private-files-strict` abstractions make sense here to mitigate this risk to some extent?

valoq:
> So wouldn't adding the `deny-sensitive-home` and `private-files-strict` abstractions make sense here to mitigate this risk to some extent?

No, these tools are commonly used to search all files, including sensitive ones. 
They also do not complain about missing permissions so blocking sensitive files will just lead to wrong outputs.

Since they do not have any other permissions that could allow extraction, full read access is fine anyway.

curiosityseeker:
> Since they do not have any other permissions that could allow extraction, full read access is fine anyway.

I understand that this is a problem. However, if it comes to confinement escape - as Alex mentioned earlier - via Px, then even reading sensitive/private files by the calling application is something we should really avoid.

valoq:
> However, they need to be handled with care: allowing something like `@{bin}/rg Px,` from other profiles would allow a confinement escape. Thus, I need to find a way to handle them correctly (possible a new linter check)

My first idea would be to define a new kind of variable that can be defined for specific permissions inside a profile and interpreted by apparmor with the result that this permission / line is ignored when there is a transition.

roddhjav:
@valoq This will exist. It is called object delegation: https://gitlab.com/apparmor/apparmor/-/wikis/AppArmorDelegation

valoq:
added ripgrep-all

roddhjav:
Merged. I will later add a linter tasks to ensure they are not used in Px transition.

PR #853, opened 2025-09-11
Update mdadm

doublez13:
There were lots of missing components of mdadm.

I have a few scripts that create and tear down MD RAID arrays.  I've ran them all and added the missing entries.

Note that mdadm has the ability to run in daemon mode and send mail when an array fails. That's why it requires all the network entries.


roddhjav:
Thanks, merged!

PR #858, opened 2025-09-12
ssh: allow ssh to authenticate to remote hosts using kerberos tickets

doublez13:
This allows the SSH client to function with kerberos auth.

Note that if the kerberos config specifies a file like `/tmp/krb5cc*` for the credential cache, then this will still fail because write access is required, whereas the abstraction only permits read access. But this should probably be fixed in a separate patch to the upstream `kerberosclient` abstraction.


roddhjav:
In this case, you need to add a write access on `/tmp/krb5cc*` directly to the profile

roddhjav:
Thanks, merged!

PR #859, opened 2025-09-12
ssh-keygen: allow execution of ssh-sk-helper

doublez13:
The ssh-sk-helper  profile was added last year but never hooked into the ssh-keygen profile.

This is needed for generating SSH keys that live on a yubikey.


roddhjav:
Thanks, merged.

PR #871, opened 2025-09-21
Create user-dirs

curiosityseeker:
Defining the user_dirs_strict variable as discussed in https://github.com/roddhjav/apparmor.d/pull/849

### DIFF
--- a/apparmor.d/tunables/multiarch.d/user-dirs
+++ b/apparmor.d/tunables/multiarch.d/user-dirs
@@ -4,7 +4,7 @@
 
 # The user directories defined below mirror the directories defined in the user-read-strict and user-write-strict abstractions. In combination with, e.g.,
 # the variables defined in .../tunables/multiarch.d/extensions they allow for rules like:
-#    owner @{user_dirs_strict/**.@{image_ext} rwkl,
+#    owner @{user_dirs_strict}/**.@{image_ext} rwkl,
 
 @{user_dirs_strict} =  @{HOME} @{HOME}/[^.]* @{MOUNTS}
 @{user_dirs_strict} += @{HOME}/@{XDG_DESKTOP_DIR} @{HOME}/@{XDG_SCREENSHOTS_DIR} @{HOME}/@{XDG_WALLPAPERS_DIR} @{MOUNTS}/@{XDG_DESKTOP_DIR}
--- a/apparmor.d/tunables/multiarch.d/user-dirs
+++ b/apparmor.d/tunables/multiarch.d/user-dirs
@@ -1,0 +1,14 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2021-2025 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+# The user directories defined below mirror the directories defined in the user-read-strict and user-write-strict abstractions. In combination with, e.g.,
+# the variables defined in .../tunables/multiarch.d/extensions they allow for rules like:
+#    owner @{user_dirs_strict/**.@{image_ext} rwkl,
+
+@{user_dirs_strict} =  @{HOME} @{HOME}/[^.]* @{MOUNTS}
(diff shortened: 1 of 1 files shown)


apparmor.d/tunables/multiarch.d/user-dirs
```diff
@@ -0,0 +1,14 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2021-2025 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+# The user directories defined below mirror the directories defined in the user-read-strict and user-write-strict abstractions. In combination with, e.g.,
+# the variables defined in .../tunables/multiarch.d/extensions they allow for rules like:
+#    owner @{user_dirs_strict/**.@{image_ext} rwkl,
```
roddhjav:
```suggestion
#    owner @{user_dirs_strict}/**.@{image_ext} rwkl,
```

roddhjav:
`@{HOME} @{HOME}/[^.]* @{MOUNTS}` cannot be there like this. Otherwise, `@{user_dirs_strict}/**.@{image_ext} rwkl,` would first resolve as `@{HOME}/**.@{image_ext} `. It  is, I think, not what you are looking for

curiosityseeker:
Ah, yes, good argument! So it's probably advisable to drop these 3 entries completely. Or do you have a better idea?

valoq:
> `@{HOME} @{HOME}/[^.]* @{MOUNTS}` cannot be there like this. Otherwise, `@{user_dirs_strict}/**.@{image_ext} rwkl,` would first resolve as `@{HOME}/**.@{image_ext} `. It is, I think, not what you are looking for

I would argue that this is what we should be looking for, as explained here https://github.com/roddhjav/apparmor.d/discussions/873

curiosityseeker:
I don't want to push ... but how should we proceed? Should we cancel this PR and go for what @valoq is suggesting? Something like:

`owner @{HOME}/[^.]**.@{image_ext} rwkl,`

possibly with the addition of the `private-files-strict` and `deny-sensitive-home` abstractions?

I'm preparing new profiles for Gwenview and Imagemagick, and I think we need a decision in order to have a consistent behavior for all affected new and existing profiles, e.g., for `okular`.

valoq:
In theory we could add both kinds of user-dirs and chose the appropriate one depending on the use case.
That said Im not sure if the more permissive approach is actually needed and if we can keep it simple, that would include less maintenance.

> possibly with the addition of the `private-files-strict` and `deny-sensitive-home` abstractions?

What for exactly? The example above `owner @{HOME}/[^.]**.@{image_ext} rwkl,` already excludes all paths in $HOME that start with a dot, so I don't see the need.

I will create a MR that implements the idea.

valoq:
@curiosityseeker I have added the new abstraction with the name user-dirs, thinking that since your implementation adds @{user_dirs_strict} it would be better if you also use the file name user-dirs-strict in case we add both approaches

curiosityseeker:
@roddhjav 
> `@{HOME} @{HOME}/[^.]* @{MOUNTS}` cannot be there like this. Otherwise, `@{user_dirs_strict}/**.@{image_ext} rwkl,` would first resolve as `@{HOME}/**.@{image_ext} `.

This does not seem to be the case. See my comment [here](https://github.com/roddhjav/apparmor.d/pull/906#issuecomment-3411893735).

valoq:
> @roddhjav
> 
> > `@{HOME} @{HOME}/[^.]* @{MOUNTS}` cannot be there like this. Otherwise, `@{user_dirs_strict}/**.@{image_ext} rwkl,` would first resolve as `@{HOME}/**.@{image_ext} `.
> 
> This does not seem to be the case. See my comment [here](https://github.com/roddhjav/apparmor.d/pull/906#issuecomment-3411893735).

The rule above results in access to hidden files. Both variables need to include [^.] to avoid this

curiosityseeker:
> The rule above results in access to hidden files. Both variables need to include [^.] to avoid this

You're right. When I tested it I stupidly forgot to comment the `private-files-strict` and `deny-sensitive-home` abstractions. Stupid! :angry: 

It does work with the file names for common extensions as suggested by you.

curiosityseeker:
I think I should close this PR in favor of @valoq 's #906 which is clearly superior. Is this okay for you?

curiosityseeker:
Closing as indicated above.

PR #878, opened 2025-09-22
lscpu: adding attach_disconnected

curiosityseeker:


roddhjav:
Thanks; as it not keep in complain mode, you don't need to update the flags file (yes, I know it is confusing, I will rethink this)

PR #880, opened 2025-09-22
fix zpool again

Stoppedpuma:
Previous fix has this error which slipped past me and apparmor_parser.

```
ERROR: Can't parse mount rule mount fstype=zfs options=(rw noatime) * -> @{MOUNTS}/,
```


roddhjav:
simply remove the '*'

PR #888, opened 2025-09-27
Fixes for many profiles

emixam16:
This commit contains fixes for many profiles

Rules in this commit were generated by running autopkgtest for all packages in Ubuntu Questing. Denials were converted to rules thanks to aa-logprof.

Rules have been generalized manually, but due to the very high amount of rules, many rules still need to be improved:

Known issues:
 - Some abstractions where added by aa-logprof, but in many cases they can be too permissive, and adding individual rules would be preferable
 - There was MANY rules to handle so I only took a small time converting and integrating these rules into profiles. All these rules need to be carefully reviewed

This commit is only a preliminary result and my goal is to integrate such a pipeline into regular package deployment pipelines so that regressions are quickly spotted.

All reviews and remarks are welcome.


emixam16:
The pipeline I used and the untreated results are available there

[test.zip](https://github.com/user-attachments/files/22573515/test.zip)

Note: I made this pipeline run on a VM with 
https://launchpad.net/~georgiag/+archive/ubuntu/apparmor.dinapparmor5/ ,  which is the ubuntu integration of this repo and not this repo directly, which can explain slight differences.

However, I tried to integrate back changes for this repo in this commit

roddhjav:
Hi @emixam16, thanks a lot, that will be massively helpful.

As you mentioned, abstraction added by aa-logprof are a bit too much (or out of scope). It is problematic as we need to understand why a rule is needed. Would you be able to send me the full raw logs? 

Also, autopkgtest tend to require special (test) paths, so they should be taken care separately. For this, I added the `--test` option to the prebuild system. Simply build the package with `just complain-test` and rules like the following will be kept:
https://github.com/roddhjav/apparmor.d/blob/c9f1471a63cb74add77db46e56c2b3d94df106cd/apparmor.d/abstractions/apt#L38-L39

I am sure I am still missing a lot of these tests only rules (I have limited resources to run autopkgtest). I think it is important to properly and separately covert these rules, to not allow too much in the base profile, and to not miss required rules (because the test version would have given a full `/tmp/**` for instance). Again, with the full logs, I could have a better idea of the need here.

Side note for later:
- All gnu core utils like `@{bin}/gnucat` now have aliases (the same will be true soon for the rust ones - not in the main branch yet). So unless the rules really is missing, they should be the same.
- `abstractions/user-tmp` is forbidden in `apparmor.d`: giving full access to `/tmp/` means you can modify someone else temporary files, it is the open door to confinement escape. That's why we spend a lot of care limiting access inside `/tmp`. But I think, most of the time, it is an autopkgtest test directory that is required.

I will continue reviewing the PR, and I some comments later.

Thanks again for the work.

apparmor.d/groups/apparmor/apparmor_parser
```diff
@@ -6,6 +6,9 @@ abi <abi/4.0>,
 
 include <tunables/global>
 
+include <abstractions/libvirt-qemu>
+include <abstractions/mapping/sshd>
+include <abstractions/user-tmp>
```
roddhjav, line 11: They are not in the profile. Also, as we both know very well what apparmor_parser does, it is very unlikely these abstractions are needed.

apparmor.d/groups/apparmor/apparmor_parser
```diff
@@ -20,6 +23,7 @@ profile apparmor_parser @{exec_path} flags=(attach_disconnected) {
   @{lib_dirs}/@{multiarch}/** mr,
   @{lib_dirs}/snapd/apparmor.d/{,**} r,
   @{lib_dirs}/snapd/apparmor/{,**} r,
+  /** r, # Can load rules from anywhere
```
roddhjav, line 26: I disagree on this. Giving access to apparmor_parser is an equivalent to giving `mac_admin` cap, therefore only allowing it to load profiles from "trusted" sources is a good thing.

However, it raises the question of development profile, I personally have:
```
  @{user_projects_dirs}/** r,
```
In my local addition to this profile. But that does not apply to everyone and thus, it should be behind a condition.

roddhjav:
The `new_rules` files (and its 9600 lines) you shared from logprof output is already a good start. There are a lot of very interesting things in it.

1. First, the `#aa:only test` I mentioned can be generalized. I already moved it to the [base abstraction](https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/abstractions/base-strict#L132), but it can be expanded to:
  ```
  #aa:only test
  @{lib}/installed-tests/ r,
  @{lib}/installed-tests/** rw,
  /usr/share/rubygems-integration/{,**} r,
  /usr/share/mutter-*/tests/{,**} r,
  /m-a/{,**} rw,
  /test/{,**} rw,
  /tmp/autopkgtest* rwlkmix,
  /tmp/autopkgtest*/ rw,
  /tmp/autopkgtest*/** rwlkmix,
  ```
  **Note:** `ix` in `/tmp/autopkgtest*/**` is required, but it could lead to confusion in some profiles as we may not know if a rule is needed by a test or not.

2. Then, a lot of rules are because of `pycompile`. They should be covered by a simple priority, such as:
  ```
  priority=1 @{lib}/@{python_name}/**/__pycache__/ w,
  priority=1 @{lib}/@{python_name}/**/__pycache__/*.pyc w,
  priority=1 @{lib}/@{python_name}/**/__pycache__/*.pyc.* w,
  ```
  **Note**: this only apply to apparmor 4.1+ so it required a bit of work in the prebuild part to not break older system.

*Only with this, `new_rules` is reduced to about 5000 lines.*

3. Then we have about 1600 rules with `/tmp/test-flatpak`, this one are a bit different (they are `{,u,re}mount` rules), we need to add them (under the `#aa:only test` directive) in the selected profiles and to ensure the other mount rules in the same profiles have the same mount options.

Finally, we end up with 3000 to 4000 lines with rules that are not tests related and that are really missing in the profiles. These must be added.

This does not apply to the abstraction proposed by logprof as most of them are fully out of scope of the profiles. But abstaction such as the `consoles` one are usually legit.

Thanks, again, for the work. This initial integration may take a bit of work, but I think quite quickly we will be able to integrate only the few new rules from canonical CI.

apparmor.d/groups/apt/apt
```diff
@@ -18,6 +18,7 @@ profile apt @{exec_path} flags=(attach_disconnected) {
   include <abstractions/nameservice-strict>
   include <abstractions/python>
   include <abstractions/ssl_certs>
+  include <abstractions/evince>
```
nobody43, line 21: Pretty big abstraction. Is it really needed fully?

apparmor.d/groups/apt/apt
```diff
@@ -71,7 +72,23 @@ profile apt @{exec_path} flags=(attach_disconnected) {
   @{bin}/test       rix,
   @{bin}/touch      rix,
 
-  @{bin}/adequate                        rPx,
+  @{tmp}/** mrix,
```
nobody43, line 75: It's a disaster [waiting to happen](https://snyk.io/blog/abusing-ubuntu-root-privilege-escalation/).

apparmor.d/groups/apt/apt-cache
```diff
@@ -12,10 +12,15 @@ profile apt-cache @{exec_path} {
...
   include <abstractions/consoles>
+  include <abstractions/evince>
 
   @{exec_path} mr,
 
   @{bin}/dpkg rPx -> child-dpkg,
+  @{bin}/gnufalse mrix,
+  @{bin}/gnutrue mrix,
+
+  @{tmp}/tmp.*/rootdir/usr/bin/dpkg mrix,
```
nobody43, line 23: Executable `/tmp`... Any user could write into this path.

apparmor.d/groups/apt/dpkg-preconfigure
```diff
@@ -72,9 +84,12 @@ profile dpkg-preconfigure @{exec_path} {
   owner /var/cache/debconf/tmp.ci/*.template.@{rand6} w,
   owner /var/cache/dictionaries-common/flag-wordlist-new w,
   owner /var/log/unattended-upgrades/unattended-upgrades-dpkg.log rw,
+  /var/cache/debconf/{config,passwords,templates}.dat{,-old,-new} r,
+  /var/cache/debconf/tmp.ci/ rw,
 
   owner @{tmp}/*.template.* rw,
   owner @{tmp}/*.config.* rwPUx,
+  owner @{tmp}/** mr,
```
nobody43, line 92: Memory mapping without execution. Is it a failed `dpkg-preconfigure
` run or just notice?

apparmor.d/groups/cups/cups-backend-implicitclass
```diff
@@ -9,12 +9,25 @@ include <tunables/global>
...
+  owner @{run}/cups/cups.sock rw,
+  owner @{run}/dbus/system_bus_socket rw,
+  owner @{bin}/gs mrpx,
+  owner /var/spool/cups/tmp/* rw,
+  @{bin}/ippfind mrix,
+  @{lib}/cups/backend/ipp px,
 
   @{exec_path} mr,
 
+  owner @{etc_ro}/nsswitch.conf r,
```
nobody43, line 25: `owner` is redundant for this path.

apparmor.d/groups/gvfs/gvfsd-metadata
```diff
@@ -33,6 +34,7 @@ profile gvfsd-metadata @{exec_path} {
   owner @{gdm_share_dirs}/gvfs-metadata/{,*} rw,
 
   owner @{HOME}/.local/ w,
+  owner @{HOME}/** rw,
```
nobody43, line 37: Combined with network access this could lead to big surprise.

apparmor.d/groups/network/netplan
```diff
@@ -12,20 +12,32 @@ profile netplan @{exec_path} flags=(attach_disconnected) {
...
   @{bin}/udevadm           rCx -> udevadm,
   @{bin}/systemctl         rCx -> systemctl,
+  @{bin}/ip mrpx,
+  @{bin}/networkctl mrpx,
+  @{bin}/nmcli mrpx,
+  @{bin}/ovs-vsctl mrix,
+  @{sbin}/ethtool mrix,
 
-  /usr/share/netplan/{,**} r,
+  /usr/share/netplan/{,**} rw,
```
nobody43, line 31: Absence of write access here is a failure? Preferably, `/usr/share` should be touched by package manager only.

apparmor.d/groups/network/openvpn
```diff
@@ -41,12 +42,14 @@ profile openvpn @{exec_path} flags=(attach_disconnected) {
...
 
   signal (receive) set=(term) peer=nm-openvpn-service,
+  signal receive set=term peer=pgrep,
 
   @{exec_path} mr,
 
   @{lib}/{,NetworkManager/}nm-openvpn-service-openvpn-helper rPx,
 
   /etc/openvpn/{,**} r,
+  owner @{etc_rw}/openvpn/{,**} rw,
```
nobody43, line 52: This is what MAC protects against: changing of configuration by potentially compromised application.

apparmor.d/groups/procps/uptime
```diff
@@ -16,6 +16,7 @@ profile uptime @{exec_path} {
   @{exec_path} mr,
 
   @{run}/systemd/sessions/@{int} r,
+  @{run}/systemd/sessions/c@{int} r,
```
nobody43, line 19:
```suggestion
  @{run}/systemd/sessions/{,@{l}}@{int} r,
```

apparmor.d/groups/systemd/systemd-hostnamed
```diff
@@ -36,6 +36,8 @@ profile systemd-hostnamed @{exec_path} flags=(attach_disconnected) {
 
   @{run}/systemd/default-hostname rw,
   @{run}/udev/data/+dmi:* r,              # for motherboard info
+  @{run}/udev/data/b*:* r,
```
nobody43, line 39: All block devices?

apparmor.d/groups/systemd/systemd-localed
```diff
@@ -29,7 +30,19 @@ profile systemd-localed @{exec_path} flags=(attach_disconnected) {
...
+  @{bin}/gnucat mrix,
+  @{bin}/gnurm mrix,
+  @{bin}/gnusort mrix,
+  @{bin}/gzip mrix,
+  @{bin}/localedef mrix,
+  @{bin}/rm mrix,
+  @{bin}/sort mrix,
+
+
+  owner @{lib}/locale/** mrwlw,
```
nobody43, line 45: Double `w`.

apparmor.d/groups/systemd/systemd-localed
```diff
@@ -52,6 +66,7 @@ profile systemd-localed @{exec_path} flags=(attach_disconnected) {
   /etc/X11/xorg.conf.d/ rw,
   /etc/X11/xorg.conf.d/.#*.conf@{hex} rw,
   /etc/X11/xorg.conf.d/*.conf rw,
+  owner @{etc_ro}/ r,
```
nobody43, line 69: Already mentioned.

apparmor.d/groups/ubuntu/apport
```diff
@@ -69,6 +80,10 @@ profile apport @{exec_path} flags=(attach_disconnected) {
         @{PROC}/@{pids}/cgroup r,
         @{PROC}/@{pids}/environ r,
         @{PROC}/@{pids}/fd/ r,
+        @{PROC}/*/[gu]id_map r,
+        @{PROC}/*/attr/current r,
+        @{PROC}/*/cmdline r,
```
nobody43, line 85: `@{pids}`?

apparmor.d/groups/utils/locale-gen
```diff
@@ -35,6 +39,8 @@ profile locale-gen @{exec_path} {
 
   /var/lib/locales/supported.d/{,**} r,
 
+  owner @{etc_ro}/** rw,
```
nobody43, line 42: Extremely broad.

apparmor.d/groups/utils/losetup
```diff
@@ -24,6 +26,9 @@ profile losetup @{exec_path} {
 
   @{sys}/block/ r,
   @{sys}/devices/virtual/block/loop@{int}/{,**} r,
+  owner @{sys}/devices/** r,
+
+  owner /var/snap/lxd/common/lxd/disks/docker-storage.img rw,
```
nobody43, line 31: Installation-specific?

apparmor.d/profiles-a-f/adequate
```diff
@@ -19,16 +19,22 @@ profile adequate @{exec_path} flags=(complain) {
   @{exec_path} r,
 
   @{sbin}/ldconfig       rix,
+  @{sbin}/ldconfig.real mrix,
+
+  @{sbin}/dhcpcd r,
 
   # It wants to ldd all binaries/libs in packages.
   @{bin}/ldd                 rCx -> ldd,
+  owner @{bin}/adequate-** r,
```
nobody43, line 28: Single wildcard?

apparmor.d/profiles-a-f/borg
```diff
@@ -39,6 +41,13 @@ profile borg @{exec_path} {
   @{bin}/fusermount{,3}         rCx -> fusermount,
   @{bin}/pass                   rPx,
   @{bin}/ssh                    rPx,
+  @{bin}/dash mrix,
+  @{bin}/gnuecho mrix,
+  @{bin}/gzip mrix,
+  @{bin}/ip mrpx,
+  @{bin}/python[23].* mrix,
+  @{sbin}/ldconfig.real mrix,
```
nobody43, line 49:
```suggestion
  @{sbin}/ldconfig{,.real} mrix,
```

apparmor.d/profiles-g-l/git
```diff
@@ -108,16 +115,22 @@ profile git @{exec_path} flags=(attach_disconnected) {
...
+    owner @{PROC}/*/fd/ r,
+
     owner @{HOME}/@{XDG_GPG_DIR}/ rw,
     owner @{HOME}/@{XDG_GPG_DIR}/** rwkl -> @{HOME}/@{XDG_GPG_DIR}/**,
 
     owner @{tmp}/.git_vtag_tmp@{rand6} r,
-
+    owner @{run}/user/@{uid}/gnupg/*/* rw,
     owner @{run}/user/@{uid}/gnupg/S.gpg-agent rw,
+    owner @{run}/user/*/gnupg/*/* rw,
```
nobody43, line 131: These three lines look very similar.

apparmor.d/profiles-g-l/ip
```diff
@@ -31,11 +33,19 @@ profile ip @{exec_path} flags=(attach_disconnected) {
   umount @{run}/netns/*,
   umount @{sys},
 
+  @{tmp}/** mrix,
```
nobody43, line 36: Completely unpredictable paths for such a mundane tools as `ip`?

apparmor.d/profiles-g-l/jekyll
```diff
@@ -20,11 +22,16 @@ profile jekyll @{exec_path} {
   @{lib}/ruby/gems/*/specifications/** r,
   @{lib}/ruby/gems/*/specifications/**.gemspec rwk,
 
+  /usr/share/rubygems-integration/*/gems/** r,
   /usr/share/rubygems-integration/*/specifications/ r,
   /usr/share/rubygems-integration/*/specifications/*.gemspec rwk,
 
   /usr/share/ruby-addressable/unicode.data r,
 
+  owner @{PROC}/*/task/@{int}/comm rw,
```
nobody43, line 31:
```suggestion
  owner @{PROC}/@{pid}/task/@{tid}/comm rw,
```

apparmor.d/profiles-m-r/initramfs-hooks
```diff
@@ -11,7 +11,9 @@ profile initramfs-hooks @{exec_path} {
   include <abstractions/base>
   include <abstractions/consoles>
   include <abstractions/fonts>
+  include <abstractions/evince>
   include <abstractions/nameservice-strict>
+  include <abstractions/evince>
```
nobody43, line 16: Duplicate.

apparmor.d/profiles-m-r/initramfs-hooks
```diff
@@ -33,6 +35,29 @@ profile initramfs-hooks @{exec_path} {
...
+  @{bin}/gnumkdir mrix,
+  @{bin}/gnumktemp mrix,
+  @{bin}/gnureadlink mrix,
+  @{bin}/gnurealpath mrix,
+  @{bin}/gnurm mrix,
+  @{bin}/gnusort mrix,
+  @{bin}/gnustat mrix,
+  @{bin}/gnutr mrix,
+  @{bin}/gnuuniq mrix,
+  @{lib}/x86_64-linux-gnu/ld-linux-*.so.* mrix,
```
nobody43, line 56:
```suggestion
  @{lib}/@{multiarch}/ld{,32,64}-*.so.* mrix,
```

apparmor.d/profiles-m-r/mkinitramfs
```diff
@@ -60,6 +62,30 @@ profile mkinitramfs @{exec_path} {
   @{lib}/dracut/dracut-install rix,
   @{sbin}/blkid     rPx,
 
+  @{tmp}/** mrix,
```
nobody43, line 65: Payload straight into initramfs.

apparmor.d/profiles-m-r/mkinitramfs
```diff
@@ -60,6 +62,30 @@ profile mkinitramfs @{exec_path} {
...
+  @{bin}/3cpio mrix,
+  @{bin}/dash ix,
```
nobody43, line 67:
```suggestion
  @{sh_path} rix,
```

apparmor.d/profiles-m-r/mumble
```diff
@@ -56,6 +58,9 @@ profile mumble @{exec_path} {
        owner @{PROC}/@{pid}/mounts r,
   deny       @{PROC}/sys/kernel/random/boot_id r,
 
+  @{lib}/ r,
+  /usr/lib64/ r,
```
nobody43, line 62: Does it fail without these rules? It's probably a regular exploration by some library, which is not necessarily needed for mumble to function.

apparmor.d/profiles-m-r/needrestart
```diff
@@ -54,6 +57,8 @@ profile needrestart @{exec_path} flags=(attach_disconnected) {
   @{run}/systemd/sessions/* r,
 
   /tmp/@{word10}/ rw,
+  /tmp/** r,
+  owner /tmp/** mr,
```
nobody43, line 61: Either is redundant.

apparmor.d/profiles-m-r/repo
```diff
@@ -63,10 +63,15 @@ profile repo @{exec_path} {
...
 
     @{bin}/gpg{,2} mr,
+    @{bin}/gpg-agent mrpx,
 
     owner @{HOME}/@{XDG_GPG_DIR}/** rwkl -> @{HOME}/@{XDG_GPG_DIR}/**,
     owner @{HOME}/.repoconfig/gnupg/** rwkl -> @{HOME}/.repoconfig/gnupg/**,
 
+    owner @{PROC}/*/fd/ r,
+
+    owner @{run}/user/*/gnupg/*/** rw,
```
nobody43, line 73:
```suggestion
    owner @{run}/user/@{uid}/gnupg/*/** rw,
```

apparmor.d/profiles-s-z/sfdisk
```diff
@@ -10,13 +10,18 @@ include <tunables/global>
...
   include <abstractions/consoles>
   include <abstractions/disks-write>
+  include <abstractions/python>
+  include <abstractions/user-tmp>
 
   capability sys_admin,
 
   @{exec_path} mr,
 
+  owner /** r,
```
nobody43, line 23: Wow. Should be user-defined tunable.

apparmor.d/profiles-s-z/unmkinitramfs
```diff
@@ -40,6 +45,9 @@ profile unmkinitramfs @{exec_path} {
         /mnt/boot/ r,
   owner /mnt/boot/initrd.img-* r,
 
+  owner @{etc_ro}/group r,
+  owner @{etc_ro}/nsswitch.conf r,
```
nobody43, line 49: `include <abstractions/nameservice-strict>`

nobody43:
This is an interesting and necessary approach. However, each rule necessity should be carefully considered - [fully automated security does not work](https://en.wikipedia.org/wiki/Automation_bias). As Alex already said, we should know *why* it's needed, at least approximately. If we don't operate tech precisely it would be magic. 
Also, I think we are reaching the point when we need for programs, including tests, to be MAC-aware. Is it theoretically possible to reach common ground with upstream?
As a general note, may I suggest:
 - `m` to non-executables should be added only if program fails without it
 - `rpx`/`rcx` should be used only after `rPx`/`rCx` fails
 - `@{PROC}` should be tested with `owner` first
 - please limit wildcard use as much as possible

In addition to `aa-logprof` and `aa-log` you could use [apparmor-suggest](https://github.com/nobody43/apparmor-suggest) for rule assessment.

emixam16:
> Hi @emixam16, thanks a lot, that will be massively helpful.

You are welcome !

> As you mentioned, abstraction added by aa-logprof are a bit too much (or out of scope). It is problematic as we need to understand why a rule is needed. Would you be able to send me the full raw logs?

I did not save the full logs, therefore, I will not be able to send them for this run of the pipeline.

However, given the amount of denials that can be generated, at least for the first runs of this pipeline, I think that relying on logs whould be extremely tedious. Generated non-merged rules/profiles are probably a better middlepoint in most cases.

IMO, there is 3 kind of abstractions:
 - The one we are sure to keep (e.g. consoles abstraction)
 - The one we are unsure about
 - The one that we clearly don't want (too broad, unrelated)

If you agree, my plan is as follows
1) Let's agree on which abstraction/profile we are sure to keep
2) I will remove all other abstraction and execute again the full pipeline with the option `--no-abstraction`
3) I will submit an updated versions of these profiles and we can decide together, which abstraction to keep/remove in which profiles.

> Also, autopkgtest tend to require special (test) paths, so they should be taken care separately. For this, I added the `--test` option to the prebuild system. Simply build the package with `just complain-test` and rules like the following will be kept:
>     * `abstractions/user-tmp` is forbidden in `apparmor.d`: giving full access to `/tmp/` means you can modify someone else temporary files, it is the open door to confinement escape. That's why we spend a lot of care limiting access inside `/tmp`. But I think, most of the time, it is an autopkgtest test directory that is required.
> [... Many other review comments ...]

At the end of the road we very often have to make a tradeoff between usability and security.

This especially applies to programs that may technically access the whole system (or a subset of it).

In some cases we may be able to restrict the allowed access to a minimum without too much risk of regression, but there is many cases where the line is thinner.

I applause the fact apparmor.d wants to keep the security as high as possible and I reckon this repo does a great job at it, but when shipping this package to distribution as we tried in Ubuntu Questing this gets more problematic.

 - If we enable apparmor.d in complain mode, dmesg logs are absolutely flooded with complain logs, making them impossible to process and masking any actual security-critical alert.
 - If we enable apparmor.d in enforce mode, the system is mostly unusable.

Now I'm not advocating for changing the security/usability tradeoff of this repository just to accommodate distributions, but I think there should be a discussion about this tradeoff with upstream, as upstream uses a different one from this repositiory.

Typically, when a binary can access all files in the system but there is a (set of) standard path(s), this repository will probably only allow this(these) path(s) when upstream may allow all possible access and restrict only sensitive access with `include <abstractions/private-files>`

I think that using two (or more) security levels for profiles, one that reduces the usage of a software to only "standard" cases to reduce the attack surface, and one that allows non-typical but possible usage of a software (e.g using a custom configuration file from a non standard path) would be the best way to keep the level of security of repository, while allowing smooth integration in distributions and therefore massively increase its usage. Now obviously, I know that this suggestion requires a fair amount of work, and I'm not asking this repository to do anything right now, I'm just braindumping ideas that could make this transition easier.

I think the most robust way to implement this in practice would be to have sets of abstractions for these two modes and/or rely on variables to know which one to use. Alternatively, although less flexible, your annotation system could be used at build-time to select a security level.

If you don't want to implement this within this repository, at least documenting when program could access more resources than currently allowed with comments within the profile, would significantly help distributions, or end-users to adapt profiles to their needs.

I know the scope of this PR is big and there is things to do in all directions, but I argue that it could be a good moment to look for these security mode as when some non-standard directories are seen in tests, this is generally evidence that the program could potentially access more (and sometimes way more) than what is currently allowed.


> This is an interesting and necessary approach. However, each rule necessity should be carefully considered - fully automated security does not work. As Alex already said, we should know why it's needed, at least approximately. If we don't operate tech precisely it would be magic.

I fully agree with you. I think that automatic tools, like this one, can be useful to detect potential missing permissions, but after that, it is crucial that to manually review in depth every single changes for the reasons you mentioned.

apparmor.d/groups/apt/apt
Vyom-Yadav: For `apt` specifically this was the output from the logs:

```
$ sudo aa_suggest.py -p apt

   apt
 [/tmp/apt-tmp-index.lo56aA d,] comm=apt-get operation=unlink
 [/tmp/apt-tmp-index.yskOBL d,] comm=apt-get operation=unlink
 [/tmp/autopkgtest.VN9DvN/build.ili/src/doc/examples/configure-index r,] comm=apt,apt-get operation=file_perm,open
 [/tmp/autopkgtest.VN9DvN/build.ili/src/test/integration/varianttable r,] comm=apt-get operation=open
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=rzK1j3I1p6
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=ysaQZPZf49
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=B4S1FZlUKu
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=NPzTmTpVhm
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=2FOAhbcOtY
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=IHzZW5NQYQ
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=mx8oBIFbK9
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=sJoMLyBQ0U
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=3PlH1hoJPR
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=GJ2UkJ0jBq
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=BYFt0w5ua4
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=KGKBKxb7Pl
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=AM0MhFrWM9
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=Ve3Gan7ulr
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=QhaNnN9qkI
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm path_diffs=vGSvLWO13X
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=V7lWn3Yido
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=6pmLA4dPAR
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=5XXtEFCeMe
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=obt5D6h6ot
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=OG8ZrrJXam
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=sO0W9av6Tc
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=9A8bMaBMeC
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=cz8sBbPca6
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=EFBa6xl1C5
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=BVzkGgmygZ
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=40VPVnoPGe
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=cPSHg7hIeY
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=cUuPTWm18x
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=TzaxoxuhvU
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=22FlGfV5kE
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=5Cq3aTN8mE
 [owner /tmp/tmp.??????????/aptconfig.conf r,] operation=file_perm,open path_diffs=gONVrKaHHe
 [owner /tmp/tmp.??????????/aptconfig.conf r,] comm=apt-get operation=file_perm,open path_diffs=QbOdRWB1kS
 [owner /tmp/tmp.??????????/downloaded/pkg-sha256-bad_1.0.tar.gz{,.??????} rwd,] comm=apt-get operation=chmod,rename_src path_diffs=5Cq3aTN8mE
 [owner /tmp/tmp.??????????/downloaded/pkg-sha256-bad_1.0.tar.gz{,.??????} wc,] comm=apt-get operation=rename_dest path_diffs=5Cq3aTN8mE,.FAILED
 [owner /tmp/tmp.??????????/downloaded/testpkg_1_all.deb w,] operation=chmod path_diffs=sO0W9av6Tc
 [owner /tmp/tmp.??????????/rootdir/etc/apt/apt.conf.d/ r,] comm=apt-get operation=file_perm,open path_diffs=vGSvLWO13X
 [owner /tmp/tmp.??????????/rootdir/etc/apt/apt.conf.d/01multiarch.conf r,] comm=apt-get operation=file_perm,open path_diffs=vGSvLWO13X
 [owner /tmp/tmp.??????????/rootdir/etc/apt/preferences.d/ r,] comm=apt-get operation=file_perm,open path_diffs=Ad9STY1m3x
 [owner /tmp/tmp.??????????/rootdir/etc/apt/sources.list.d/ r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/etc/apt/sources.list.d/apt-test-sid-deb-src.list r,] comm=apt-get operation=file_perm path_diffs=SyjBJeOa84
 [owner /tmp/tmp.??????????/rootdir/etc/apt/sources.list.d/apt-test-sid-deb.list r,] comm=apt-get operation=file_perm,open path_diffs=SyjBJeOa84
 [owner /tmp/tmp.??????????/rootdir/etc/apt/sources.list.d/apt-test-unstable-deb-src.list r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/ r,] comm=apt-get operation=file_perm,open path_diffs=iKWDUaFLzR
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/pkgcache.bin d,] comm=apt-get operation=unlink path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/pkgcache.bin wc,] comm=apt-get operation=rename_dest path_diffs=DB202rnH4O
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/pkgcache.bin d,] comm=apt-get operation=unlink path_diffs=iKWDUaFLzR
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/pkgcache.bin.nZvd3N rwdc,] comm=apt-get operation=chmod,file_perm,mknod,open,rename_src path_diffs=DB202rnH4O
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin wc,] operation=rename_dest path_diffs=gfKZm9rfvB
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin wc,] comm=apt-get operation=rename_dest path_diffs=DB202rnH4O
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin d,] comm=apt-get operation=unlink path_diffs=iKWDUaFLzR
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin.BmJd7m rwc,] comm=apt-get operation=mknod,open path_diffs=iKWDUaFLzR
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin.D0jnw8 rwd,] comm=apt-get operation=rename_src path_diffs=DB202rnH4O
 [owner /tmp/tmp.??????????/rootdir/var/cache/apt/srcpkgcache.bin.TNUqMg rwdc,] operation=chmod,file_perm,mknod,open,rename_src path_diffs=gfKZm9rfvB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states r,] comm=apt-get operation=file_perm,open path_diffs=bIJpTkCkgg
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states r,] comm=apt-get operation=file_perm,open path_diffs=LOv1yM3Sop
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states r,] comm=apt-get operation=file_perm,open path_diffs=7PKN50Mq9J
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states.IHWu2k rwdc,] comm=apt-get operation=chmod,mknod,open,rename_src path_diffs=bIJpTkCkgg
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states.JfOKQQ rwdc,] comm=apt-get operation=chmod,mknod,open,rename_src path_diffs=7PKN50Mq9J
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/extended_states.NXENyG rwdc,] comm=apt-get operation=chmod,mknod,open,rename_src path_diffs=LOv1yM3Sop
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/ r,] comm=apt-get operation=file_perm,open path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/ r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.6Wg2fWaQ3a_aptarchive_dists_stable_main_binary-amd64_Packages r,] comm=apt-get operation=file_perm,open path_diffs=6Wg2fWaQ3a
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.6Wg2fWaQ3a_aptarchive_dists_stable_main_binary-armel_Packages r,] comm=apt-get operation=file_perm,open path_diffs=6Wg2fWaQ3a
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.6Wg2fWaQ3a_aptarchive_dists_stable_main_i18n_Translation-en r,] comm=apt-get operation=file_perm,open path_diffs=6Wg2fWaQ3a
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.Ad9STY1m3x_aptarchive_dists_unstable_main_binary-all_Packages r,] comm=apt-get operation=file_perm,open path_diffs=Ad9STY1m3x
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.Ad9STY1m3x_aptarchive_dists_unstable_main_binary-i386_Packages r,] comm=apt-get operation=open path_diffs=Ad9STY1m3x
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.Ad9STY1m3x_aptarchive_dists_unstable_main_i18n_Translation-en r,] comm=apt-get operation=file_perm,open path_diffs=Ad9STY1m3x
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.SyjBJeOa84_aptarchive_dists_sid_InRelease r,] comm=apt-get operation=file_perm,open path_diffs=SyjBJeOa84
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.Vk1iXKwluF_aptarchive_dists_unstable_main_binary-amd64_Packages r,] comm=apt-get operation=file_perm path_diffs=Vk1iXKwluF
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/_tmp_tmp.uhtKlAokzv_aptarchive_dists_unstable_main_binary-all_Packages r,] comm=apt-get operation=file_perm path_diffs=uhtKlAokzv
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:39625_dists_unstable_InRelease r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:39625_dists_unstable_main_binary-i386_Packages.lz4 r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:39625_dists_unstable_main_i18n_Translation-en.lz4 r,] comm=apt-get operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:39817_Packages wc,] comm=apt-get operation=rename_dest path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:40289_InRelease r,] comm=apt-get operation=file_perm,open path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:45695_InRelease r,] operation=open path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/localhost:46143_dists_unstable_Release rw,] comm=apt-get operation=chmod,file_perm,open path_diffs=I9NKPISzGc
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/ r,] comm=apt-get operation=file_perm,open path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:39119_dists_unstable_main_source_Sources.gz w,] comm=apt-get operation=chmod path_diffs=vGSvLWO13X
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:39625_dists_unstable_main_binary-i386_Packages.lz4 w,] comm=apt-get operation=chmod path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:39817_Packages rwd,] comm=apt-get operation=rename_src path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:39817_Packages.zst d,] comm=apt-get operation=unlink path_diffs=WwWcU6VvdB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40289_InRelease rw,] comm=apt-get operation=chmod,file_perm,open path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40289_Packages w,] comm=apt-get operation=chmod path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40289_Packages.gz w,] comm=apt-get operation=chmod path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40289_Release w,] comm=apt-get operation=chmod path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40289_Release.gpg w,] comm=apt-get operation=chmod path_diffs=31OF1231WP
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:40567_dists_unstable_main_binary-i386_Packages.gz w,] comm=apt-get operation=chmod path_diffs=KEsQhf2C17
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:42671_redirectme_dists_unstable_main_source_Sources.gz w,] comm=apt-get operation=chmod path_diffs=AoRs61TRnj
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:43177_dists_unstable_main_source_Sources.gz w,] operation=chmod path_diffs=REdljKd1HV
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:43799_dists_unstable_main_source_Sources.xz w,] comm=apt-get operation=chmod path_diffs=a2vY79KqoM
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:44805_dists_unstable_main_binary-amd64_Packages.gz w,] comm=apt-get operation=chmod path_diffs=3PlH1hoJPR
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:44805_dists_unstable_main_source_Sources.gz w,] comm=apt-get operation=chmod path_diffs=3PlH1hoJPR
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:45695_InRelease rw,] operation=chmod,file_perm,open path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:45695_Packages w,] operation=chmod path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:45695_Packages.gz w,] operation=chmod path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:45695_redirectme_Packages.diff_2025-10-06-1201.08.lz4 rwd,] operation=chmod,rename_src path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/apt/lists/partial/localhost:45695_redirectme_Packages.lz4.ed wc,] operation=rename_dest path_diffs=s2RZwUm7AB
 [owner /tmp/tmp.??????????/rootdir/var/lib/dpkg/lock-frontend rwck,] comm=apt-get operation=file_lock,open path_diffs=SyjBJeOa84
 [owner /tmp/tmp.??????????/rootdir/var/lib/dpkg/status r,] operation=file_perm,open path_diffs=gfKZm9rfvB
 [owner /tmp/tmp.??????????/rootdir/var/lib/dpkg/status r,] comm=apt-get operation=file_perm,open path_diffs=6Wg2fWaQ3a
 [owner /tmp/tmp.??????????/rootdir/var/lib/dpkg/status r,] comm=apt-get operation=file_perm,open path_diffs=Ad9STY1m3x
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=TpwVmc1HAq
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=bIJpTkCkgg
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=TLUUe66HLs
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=ysaQZPZf49
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=LOv1yM3Sop
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=B4S1FZlUKu
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=2FOAhbcOtY
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=7PKN50Mq9J
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] operation=file_perm path_diffs=obt5D6h6ot
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] operation=file_perm path_diffs=L6loPI7Vq7
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=9A8bMaBMeC
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=pUMDc4Vhsl
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=BVzkGgmygZ
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=cPSHg7hIeY
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz dc,] operation=mknod,unlink path_diffs=R0LxScDFLi
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=05PryLWk0O
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] comm=apt-get operation=file_perm path_diffs=22FlGfV5kE
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/eipp.log.xz w,] operation=file_perm path_diffs=gONVrKaHHe
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/history.log wc,] comm=apt-get operation=chmod,file_perm,open path_diffs=pUMDc4Vhsl
 [owner /tmp/tmp.??????????/rootdir/var/log/apt/term.log wc,] comm=apt-get operation=chmod,file_perm,open path_diffs=pUMDc4Vhsl
 [/usr/share/dpkg/varianttable r,] comm=apt,apt-get operation=file_perm,open
 [/{,usr/}bin/gnufalse mrx,] comm=apt-get,false operation=exec,file_mmap path_diffs=usr/
 [@{run}/systemd/inhibit/*.ref w,] comm=apt-get operation=file_receive path_diffs=/run,1
 [@{run}/systemd/inhibit/*.ref w,] operation=file_receive path_diffs=/run,4
 [@{run}/systemd/inhibit/*.ref w,] operation=file_receive path_diffs=/run,5
 [@{run}/systemd/inhibit/*.ref w,] operation=file_receive path_diffs=/run,6

   apt//dpkg-source
 [/etc/dpkg/origins/ubuntu r,] comm=dpkg-source operation=file_perm,open
 [owner /tmp/tmp.??????????/downloaded/testpkg-1.0/ c,] comm=dpkg-source operation=mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.KkfJi/ wc,] comm=dpkg-source operation=chmod,mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.WxhyX/ wc,] comm=dpkg-source operation=chmod,mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.WxhyX/testpkg-1.0/ c,] comm=tar operation=mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.ZvEuf/ wc,] comm=dpkg-source operation=chmod,mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.xqjxS/ wc,] comm=dpkg-source operation=chmod,mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/downloaded/testpkg_1.0.tar.xz.tmp-extract.xqjxS/testpkg-1.0/ c,] comm=tar operation=mkdir path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/incoming/testpkg_1.0.dsc r,] comm=dpkg-source operation=file_perm,open path_diffs=2ppKcyfN8h
 [owner /tmp/tmp.??????????/incoming/testpkg_1.0.tar.xz r,] comm=dpkg-source,unxz operation=file_perm,open path_diffs=2ppKcyfN8h
 [/{,usr/}bin/gnurm mr,] comm=rm operation=file_mmap path_diffs=usr/

   apt//dpkg-source▶gnurm
 [/{,usr/}bin/gnurm r,] comm=rm operation=file_mmap path_diffs=usr/

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶dpkg
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶dpkg▶dpkg-deb
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶dpkg▶dpkg-deb▶tar
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶dpkg▶gnurm
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶faked-sysv
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶getopt
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶gnucut
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.gplsTTiNxT/rootdir/usr/bin/dpkg▶fakeroot-sysv▶sed
 [owner /dev/pts/[0-9]* rw,] comm=dpkg-split operation=file_inherit path_diffs=1

   apt▶/tmp/tmp.hgHUqtmpae/rootdir/usr/bin/dpkg▶fakeroot-sysv
 [/{,usr/}bin/faked-sysv rx,] comm=fakeroot operation=exec path_diffs=usr/
 [UNKNOWN_RULE]               comm=sed name=/usr/bin/sed operation=file_mprotect requested_mask=r
 [UNKNOWN_RULE]               comm=sed name=/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 operation=file_mprotect requested_mask=r

   apt▶/tmp/tmp.j8H6lrlEb2/pre-install-pkgs-v3.sh
 [owner /tmp/tmp.??????????/pre-install-pkgs-v3.list a,] comm=grep operation=file_inherit path_diffs=j8H6lrlEb2
 [/{,usr/}bin/{,e,f}grep mrix,] comm=grep,grep,pre-install-pkg operation=exec,file_mmap path_diffs=usr/

   apt▶/tmp/tmp.j8H6lrlEb2/rootdir/usr/bin/dpkg▶fakeroot-sysv
 [/{,usr/}bin/gnucut r,] comm=fakeroot operation=getattr path_diffs=usr/
 [UNKNOWN_RULE]               comm=cut name=/usr/bin/gnucut operation=file_mprotect requested_mask=r
 [UNKNOWN_RULE]               comm=cut name=/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 operation=file_mprotect requested_mask=r

   apt▶/tmp/tmp.j8H6lrlEb2/rootdir/usr/bin/dpkg▶fakeroot-sysv▶gnucut
 [UNKNOWN_RULE]               comm=cut name=/usr/lib/x86_64-linux-gnu/libc.so.6 operation=file_mprotect requested_mask=r

   apt▶gnufalse
 [/{,usr/}bin/gnufalse r,] comm=false operation=file_mmap path_diffs=usr/

This distro is not supported. Watch out for inconsistencies.
```

Based on them, it would translate to rules like:

```
owner /tmp/tmp.*/aptconfig.conf r,
owner /tmp/tmp.*/rootdir/etc/apt/apt.conf.d/ r,
owner /tmp/tmp.*/rootdir/etc/apt/apt.conf.d/* r,
owner /tmp/tmp.*/rootdir/etc/apt/preferences.d/ r,
owner /tmp/tmp.*/rootdir/etc/apt/sources.list.d/ r,
owner /tmp/tmp.*/rootdir/etc/apt/sources.list.d/* r,
owner /tmp/tmp.*/rootdir/var/cache/apt/ r,
owner /tmp/tmp.*/rootdir/var/cache/apt/* rwdc,
owner /tmp/tmp.*/rootdir/var/lib/apt/extended_states r,
owner /tmp/tmp.*/rootdir/var/lib/apt/extended_states.* rwdc,
owner /tmp/tmp.*/rootdir/var/lib/apt/lists/ r,
owner /tmp/tmp.*/rootdir/var/lib/apt/lists/* r,
owner /tmp/tmp.*/rootdir/var/lib/apt/lists/partial/ r,
owner /tmp/tmp.*/rootdir/var/lib/apt/lists/partial/* rwdc,
owner /tmp/tmp.*/rootdir/var/lib/dpkg/lock-frontend rwck,
owner /tmp/tmp.*/rootdir/var/lib/dpkg/status r,
owner /tmp/tmp.*/rootdir/var/log/apt/* rwdc,
owner /tmp/tmp.*/downloaded/* rwc,
owner /tmp/tmp.*/incoming/* r,
owner /tmp/tmp.*/pre-install-pkgs-v3.list a,
owner /tmp/apt-tmp-index.* d,
```

But I don't think these directories are created by apt during regular execution. I think these are temporary directories created for testing. I investigated the apt testing source code a bit and it also indicates towards that:

```bash
getaptconfig() {
	if [ -f ./aptconfig.conf ]; then
		echo "$(readlink -f ./aptconfig.conf)"
	elif [ -f ../aptconfig.conf ]; then
		echo "$(readlink -f ../aptconfig.conf)"
	elif [ -f ../../aptconfig.conf ]; then
		echo "$(readlink -f ../../aptconfig.conf)"
	elif [ -f "${TMPWORKINGDIRECTORY}/aptconfig.conf" ]; then
		echo "$(readlink -f "${TMPWORKINGDIRECTORY}/aptconfig.conf")"
	fi
}
```

So, I don't think it requires this permission.

Vyom-Yadav: I didn't find any `m` or `x` requested operations for the `/tmp` directory in the logs. So I don't think that's required.

apparmor.d/groups/apt/apt-cache
Vyom-Yadav: `tmp.*/rootdir/usr/bin/dpkg mrix,` I'm inclined to believe this is something apt does for testing and not related to how this application functions outside of that. 

https://codesearch.debian.net/search?q=getaptconfig+package%3A%5CQapt%5CE&literal=1
https://codesearch.debian.net/search?q=APT_CONFIG+package%3A%5CQapt%5CE&literal=1

apparmor.d/groups/apt/apt
```diff
@@ -71,7 +72,23 @@ profile apt @{exec_path} flags=(attach_disconnected) {
...
+  @{bin}/apt-get px,
+  @{bin}/dash ix,
+  @{bin}/gnucat mrix,
+  @{bin}/gnudd mrix,
+  @{bin}/gnudf mrix,
+  @{bin}/gnuecho mrix,
+  @{bin}/gnufalse mrix,
+  @{bin}/gnuhead mrix,
+  @{bin}/gnuid mrix,
+  @{bin}/gnutest mrix,
```
Vyom-Yadav, line 85: Not sure where all of these are coming from, but `gnufalse` is being used for just testing:

```
 $ grep -r "gnufalse\|/bin/false" .                                    
./test/integration/framework:	echo 'Dir::Bin::ischroot "/bin/false";' >> aptconfig.conf
```

I think others might be used for the same. But needs verification for every statement.

roddhjav: This part is a bit problematic; giving access to files like the following only in test mode is fine.

https://github.com/roddhjav/apparmor.d/blob/c9f1471a63cb74add77db46e56c2b3d94df106cd/apparmor.d/abstractions/apt#L38-L39

Having to handle `ix` rule of normal binary means that the test script need to be fully profiled too.

apparmor.d/groups/apt/apt
nobody43: Thanks for the evaluation, and testing.
I wonder, is it possible to substitute `/tmp` with `/run/user/$uid` as temporary directory in tests? `/tmp` is really outdated by modern security standards.
BTW, you should use:
`sudo aa_suggest.py -S AppArmor.d -c -p apt`
`-S AppArmor.d` for proper tunables for *this* project (non-exhaustive) and `-c` for actually usable permissions (not what kernel reports)

apparmor.d/groups/apt/apt
Vyom-Yadav: `dd` seems to be used in a script: 

```
 $ grep -r "dd if"                         
debian/apt.apt-compat.cron.daily:	RANDOM=$(( $(dd if=/dev/urandom bs=2 count=1 2> /dev/null | cksum | cut -d' ' -f1) % 32767 ))
test/integration/test-apt-update-expected-size:	dd if=/dev/zero of=aptarchive/dists/unstable/InRelease bs=1M count=2 2>/dev/null
test/integration/test-acquire-same-file-multiple-times:	dd if="$TESTFILE" of=./downloaded/foo1 bs=500 count=1 2>/dev/null
test/integration/test-acquire-same-file-multiple-times:	dd if="$TESTFILE" of=./downloaded/foo2 bs=500 count=1 2>/dev/null
test/integration/test-acquire-same-file-multiple-times:	dd if="$TESTFILE" of=./downloaded/foo1 bs=500 count=1 2>/dev/null
test/integration/test-acquire-same-file-multiple-times:	dd if="$TESTFILE" of=./downloaded/foo2 bs=500 count=1 2>/dev/null
test/integration/test-partial-file-support:	dd if="$1" bs=1 count="$2" of="$3" 2>/dev/null
test/integration/test-bug-lp1445239-download-loop:dd if=/dev/zero of=$TESTFILE bs=100k count=1 2>/dev/null
test/integration/test-bug-lp1445239-download-loop:dd if=/dev/zero of=$TARGET bs=99k count=1 2>/dev/null
test/integration/framework:		testsuccess dd if=/dev/zero of="${1}" bs="$(stat -c %s "${1}.bak")" count=1
test/integration/skip-apt-download-progress:testsuccess dd if=/dev/zero of=./aptarchive/$TESTFILE bs=1600k count=1
```

`debian/apt.apt-compat.cron.daily:	RANDOM=$(( $(dd if=/dev/urandom bs=2 count=1 2> /dev/null | cksum | cut -d' ' -f1) % 32767 ))`

roddhjav:
> However, given the amount of denials that can be generated, at least for the first runs of this pipeline, I think that relying on logs whould be extremely tedious. Generated non-merged rules/profiles are probably a better middlepoint in most cases.

I am not planning to handle the logs manually (although some patterns are easy to spot). I have `aa-log` that can generate rules and give me an overview of what is required. The main point here is to make differences between what is needed only for tests and what is really missing. As `aa-log` is stupid, it cannot propose abstraction, which is a plus in this case. It is also aware of variable and guideline in this project, which helps a lot to add new rules.

Also, I am not expecting to fix all issues within this first PR, and with only one run of the test suite. This need to be done step by step: first focusing on test only rules, then after, when they are fewer logs, looking at the other rules.

> If you agree, my plan is as follows

> Let's agree on which abstraction/profile we are sure to keep
> I will remove all other abstraction and execute again the full pipeline with the option --no-abstraction
>I will submit an updated versions of these profiles and we can decide together, which abstraction to keep/remove in which profiles.

The plan looks good. As you said, I think you need to fully re-run the full autopkgtest pipeline, especially as some of the issues should now be fixed. Thanks to the test script you provided, I was able to run some part of it myself, in order to check the among of rule that are needed. The result is in https://github.com/roddhjav/apparmor.d/pull/893. I think it should handle most of the test only rules. Including the rules raised by @Vyom-Yadav 

However, please wait for https://github.com/roddhjav/apparmor.d/pull/893 to be merged (as I am currently busy holidaying, not before next week) before running it again. It also supports new gnu/rust utils path, so it is a must have on Ubuntu 25.10. Also, ensure you build the profiles with `just complain-test` otherwise the tests only rule would not be enabled.

> The one we are sure to keep (e.g. consoles abstraction)
> The one we are unsure about
> The one that we clearly don't want (too broad, unrelated)

In general, the consoles abstraction is safe to keep. Most capabilities should be legitimate. However, sorry, but all other abstraction seems too wide or simply out of scope (`<abstractions/desktop>`  in `udisksd`, `<abstractions/bwrap>` in `dbus-session`).

Things like `/tmp/**` or `/dev/shm/**` are however forbidden (even in less secure mode). Rules like `/**` is limited to weakly confined programs (diff, gcc...), weakly confined profile cannot be transitioned from another profile. Package manager are valid exception.

There are also case, with some recommended way to handle things. For example:
- Some program should always be used in a subprofile using a provided abstraction: https://apparmor.pujol.io/development/abstractions/#context-helper
- `<abstractions/gio-open>`: should be handled with a [child-open](https://apparmor.pujol.io/development/internal/#open-resources) profile instead of the full (large) abstraction.
- Always use variables, and try to respect [the guidelines](https://apparmor.pujol.io/development/guidelines/): `/usr/libexec` does not exist on some distributions...
- Things like `@{bin}/foo mrix,` or `@{bin}/foo mrpx,`, should be `@{bin}/foo ix,` or `@{bin}/foo Px,` in most cases (except interpreter).
- When there is a profile, `Px` is usually recommended. E.g. in `cups-backend-implicitclass` use `@{bin}/ippfind Px,` instead of `@{bin}/ippfind ix,`. It also reduce the amount of rule needed in `cups-backend-implicitclass`.
- You cannot use dbus abstraction from apparmor here (they would conflict with peer label). E.g. use `bus-session` instead of `dbus-session`. See https://apparmor.pujol.io/development/dbus/
- Do not use `abstractions/nameservice`, even when network is required (I am trying to keep consistency across profiles).

If you have any doubt, do not hesitate to share the full logs with me.

> I know the scope of this PR is big and there is things to do in all directions, 

I agree; we do not need to fix all issue in one PR; you can create as many PR as you want, they will be easier to review.

> If we enable apparmor.d in complain mode, dmesg logs are absolutely flooded with complain logs, making them impossible to process and masking any actual security-critical alert.

I may be used to this thing more, or I am still lacking the full picture; however with the initial test integration, I do not see that many logs. I did not run the full test suite yet, but, it seems to me that tests specific path are at the origin of 75% of the logs.

> If we enable apparmor.d in enforce mode, the system is mostly unusable.

Currently, 90% of the issues on Ubuntu are dbus related:
1. [Despite the massive amount of work done specifically on dbus](https://apparmor.pujol.io/development/dbus/), a lot of rules are still needed, and this will break on enforce mode. I am currently in the process of restructuring all dbus interfaces abstraction, but it is a long work. Note: because only Ubuntu support dbus mediation it clearly does not help: most of the users (and bug report) of this project are for arch based distribution.
2. Snap app are broken unless you use this patch: https://github.com/canonical/snapd/compare/release/2.71...roddhjav:snapd:release/2.71 It is currently a wip: https://github.com/canonical/snapd/pull/15900

I sincerely hope that the other 10% will be fixed by the integration of autopkgtest test suite.

**Note 1:** As I use `apparmor.d` in complain mode on Ubuntu 24.10 on my work computer, and have [test VM for Ubuntu 25.10](https://apparmor.pujol.io/development/vm/), I am very well aware of the issues.

**Note 2:** Looking back at the last two weeks of this project, we also have sustainability issues. In two weeks:
- Ubuntu 25.10 added gnu path, and rust utils path; breaking all profiles;
- Gnome 49 added glycin loader in sandbox, breaking all profiles;

Nothing major, however without serious support and maintenance work from companies *(read money)*, this project will die.

> At the end of the road we very often have to make a tradeoff between usability and security.

Yes, I fully agree (out of scope of this PR). My plan is to integrate multiple level of security implementation with the support of apparmor condition. I still need to define the levels (possibly **a lot**) and they should also integrate into thing that would be the apparmor equivalent of the hundred of selinux booleans.

We can define them in a separate discussion.

**Meanwhile, the most paranoid version of the rules should still pass the test suite.** Thus, it does not fully apply here.

However, while this project is still not stable; it could be interesting to provide a "prod ready" deployment that explicitly relax everything that is known to cause issues (such as dbus). But again the base system has to be the most paranoid one for development reason.

apparmor.d/groups/apt/apt
roddhjav: `debian/apt.apt-compat.cron.daily` is covered with https://github.com/roddhjav/apparmor.d/blob/main/apparmor.d/groups/cron/cron-apt-compat

The use of dd in `test/integration/` will be more problematic to handle.

emixam16:
> However, please wait for https://github.com/roddhjav/apparmor.d/pull/893 to be merged

Sure, I'll wait for that, no issues :)

> I agree; we do not need to fix all issue in one PR; you can create as many PR as you want, they will be easier to review.

Sure, I'll split it up in the next run to make the review easier.

> If you have any doubt, do not hesitate to share the full logs with me.

I'll save AppArmor logs in the next pipeline run. There might be some valuable insight there.

> Without serious support and maintenance work from companies (read money), this project will die.

It is indeed hard to maintain such a huge project, given how fast the Open Source environment evolves.

Canonical will increase its effort to support apparmor.d. As we aim to support apparmor.d in Ubuntu 26.04 LTS, we will try to have members of the broader security team to review some profiles in this repository.

Additionally, in the longer term: if we manage to enable such a pipeline into packaging CI in a distro-agnostic way,  it could save time and automatically detect most regressions. I think that would simplify profiles' maintenance. I will work in this direction.

roddhjav:
#893 is now merged, fell free to run the test again.

Thanks to the script you shared, I was able to almost run the full test suite, so most of the test integration should be already in place now. See https://github.com/roddhjav/apparmor.d/tree/main/tests/autopkgtest. The full integration is detailed under https://apparmor.pujol.io/development/autopkgtest/.

I suppose you will use canonical VM and not the ones in this project to run them. So here are some points:

- To enable tests only directory, you need to build the package in test mode: `just complain-test` (it is automatic when you do `just img ubuntu 25.10 test`, but that is internal apparmor.d dev things).

- The package needs to be part of the VM image. If you install them just before the tests, it will lead to issues. In your `setup-commands`, you have `sudo cp -r $VM_DEST/* /etc/apparmor.d/ || true; sudo apparmor_parser -r /etc/apparmor.d/;`. It is problematic as all the program already running won't be confined under their expected profile. Also, it is slow as it is run on VM start up, ie: before each pkg test.

- In the VM image, before installing the package, enable early policy load. Some profiles won't work without it.

- Use the package as it also handle some profile conflict across apparmor version that will break things if you copy and paste the apparmor.d directory (eg: profile defined twice...).

They are still a bit of work and tests to be made, but we don't have that many logs any more, and the result kind of looks good to me to be honest.

PR #891, opened 2025-09-29
Add allowed paths for correct generation of swap target

myrslint:
The generator for GPT mounts also creates a swap target, when a swap partition is available. Write access to paths relating to this target was missing. They were added in this commit.


roddhjav:
Thanks; merged!

PR #906, opened 2025-10-16
add user-files variable

valoq:
This adds a user-dirs abstraction as discussed in #871 

It also adds a profile for magic as an example use case, allowing to read and write only files with the image_ext extension

Reasoning: https://github.com/roddhjav/apparmor.d/discussions/873

Updates:
To allow cleaner and less error prone profiles, the files variables were also added.


valoq:
We may need a rule check to ensure that  @{user_dirs}/ is never followed by ** in any profile since it will overwrite the [^.] restriction. 

There should also be a check for `@{HOME}/** rw,` which currently exists in some profiles like thunar, nemo, nautilus, gvfsd-trash, gvfsd-mtp

valoq:
To prevent writing to `$HOME/.file.extension` (though I dont yet see any issue with that) the only solutions I see are either using `@{user_dirs}.@{image_ext} rw,` or to use
`@{user_dirs}/[^.]*@{image_ext} rw, `in every profile

Perhaps we should add another variable called image_files that takes care of this and is defined as` image_files = [^.]*@{image_ext}`
That way the profiles can be written as: `@{user_dirs}/@{image_files} rw,` which looks much cleaner and is less error prone

Or we add image_files without blocking hidden files since there is probably nothing wring with writing to `$HOME/.bashrc.png` considering that controls files in home do not have file extensions?

curiosityseeker:
The example in `user-dirs` should be modified ;-)

Regarding `@{user_dirs} =  @{HOME} @{HOME}/[^.]**` : this seems to work despite despite what @roddhjav mentioned  [here](https://github.com/roddhjav/apparmor.d/pull/871#issuecomment-3316084316).

I created the directory `@{HOME}/.test` and was **not** able to open an image therein (neither in other dot directories) with my gwenview profile using the rule `owner @{user_dirs}/*.@{image_ext} rwl,` (but everywhere else in @{HOME}). 

However, I got error messages that the respective directories could not be opened. So I had to add the rule: `owner @{HOME}/[^.]**/ r,` to fix this issue.

EDIT: see my correction [here](https://github.com/roddhjav/apparmor.d/pull/871#issuecomment-3418647807).

valoq:
@curiosityseeker please try again with the current version.

I added the files variables to allow much cleaner profiles that do not require the explicit deny rules for hidden files.

As a side effect you also can't write files to `$HOME/Documents/.test.png` if you use the image files variable. I think the only common use case for this would be text files, for which I added an extra variable to allow it, though it needs to be used carefully just like any rules allowing access to hidden files.

curiosityseeker:
I tested the new version and it seems to work as advertised! :+1: 

Perhaps it should be added to `user-dirs` that the rule `owner @{user_dirs}/**/ r,` is needed to prevent error messages when opening directories. Dot directories cannot be opened with that rule.

curiosityseeker:
I suggest that you add

`@{user_dirs}/[^.]*.[pP][dD][fF] rw,`

to the `magick` profile as it can convert images to pdf and vice versa.

curiosityseeker:
Shouldn't we add `owner` to `@{user_dirs}/@{image_files} rw,` and `@{user_dirs}/[^.]*.[pP][dD][fF] rw,`?

valoq:
> Shouldn't we add `owner` to `@{user_dirs}/@{image_files} rw,` and `@{user_dirs}/[^.]*.[pP][dD][fF] rw,`?

I was considering this, however I am not sure if it makes more sense.
Including mounts in user_dirs means shared directories like NAS drives are permitted as well and files there are not necessarily owned by the user. Additionally the restriction does not follow any clear threat model that would require it. It makes sense to allow some file types only for the owner, such as key files or other sensitive resources, but the files in the added files variables are typical user files. Since the user running the application still requires the DAC access permissions, this additional layer just seems more in the way of the user.

With the extensions and file type variables I think we can take a big step towards the android model and restrict many profiles further to allow access only to relevant file types, while avoiding most restrictions based on the file path. This reduces the chance of over-blocking while providing more effective isolation at the same time. Adding `owner` would be a step back and add more risks of over-blocking. I would rather keep the `owner` attribute to more relevant use cases like specifying configurations files for the application.

roddhjav:
See https://github.com/roddhjav/apparmor.d/discussions/873

curiosityseeker:
Unfortunately I have to report that this approach does not seem to work reliably at least for KDE applications. 

I've tested it with my gwenview profile. It looks like this:

```
abi <abi/4.0>,

include <tunables/global>

@{image_files} = [^.]*.@{image_ext}
@{user_dirs} =  @{HOME} @{HOME}/[^.]** @{MOUNTS} @{MOUNTS}/[^.]**

@{exec_path} = @{bin}/gwenview
profile gwenview @{exec_path} {
  include <abstractions/base-strict>
  # include <abstractions/deny-sensitive-home>
  include <abstractions/disks-read>
  include <abstractions/graphics>
  include <abstractions/kde-strict>
  include <abstractions/oneapi>
  # include <abstractions/private-files-strict>
  include <abstractions/thumbnails-cache-write>

....

  @{exec_path} mr,

  @{lib}/kf6/kioworker Px,

...

 owner @{user_dirs}{,/[^.]**/} r,
 owner @{user_dirs}/@{image_files} rwl,

....
```
If I start Gwenview and open @{HOME] with File -> Open , dot files are not displayed and dot directories cannot be opened (with the exception of ~/.config and ~/.cache as there are extra rules ion the profile) - good!

However, there are, e.g., some pdf files in the list and I was able to open them! Even worse, there are some odt files from LibreOffice, and they are obviously handled as archives by KDE which I could open - it's shown in the command line as `krarc:///home/XXX/ZZZ.odt`. I was able to traverse to a "Thumbnails" sub-directory within that archive which contains a `thumbnail.png` showing the content of the document as a mini-picture. Ouch! 

Note that I was using the profile in enforce mode! I cannot really explain this behavior. It seems that via the `kde-strict` abstraction (which contains `kde-base`) and/or kioworker some other KDE components come into play which undermine the restrictions defined by the rules. The question is if this behaviour can be mitigated somehow.

And another question is if other DEs are also affected.

Perhaps @cboltz can share some insights?

valoq:
Can you share the complete gwenview profile? I did not find it in the repo.
My suspicion is that it includes something like a transition to xdg-open which expands the permissions again or communicates through the xdg-desktop-portal.

I tested several profiles again and it seems to work as expected for me.
For example, if you take a simple image viewer like imv and allow it to open only png files, it cant open other image types.


```
abi <abi/4.0>,
include <tunables/global>

@{exec_path} = @{bin}/imv-wayland
@{att} = ""
profile imv /{,usr/}bin/imv-wayland {
  include <abstractions/base-strict>
  include <abstractions/desktop-restricted>
  include <abstractions/fontconfig-cache-read>
  include <abstractions/graphics>
  ##include <abstractions/user-read-strict>

  @{exec_path} mr,
  /etc/imv_config r,

  @{user_dirs}/[^.]*.[pP][nN][gG] r,

  owner @{user_config_dirs}/imv/config r,
  owner @{run}/user/@{uid}/imv-*.sock w,
  include if exists <local/imv>
}

# vim:syntax=apparmor
```

curiosityseeker:
This is the complete gwenview profile I'm testing right now:

```
# Copyright (C) 2025 You <your@email>

abi <abi/4.0>,

include <tunables/global>

@{image_files} = [^.]*.@{image_ext}
@{user_dirs} =  @{HOME} @{HOME}/[^.]** @{MOUNTS} @{MOUNTS}/[^.]**

@{exec_path} = @{bin}/gwenview
@{att} = ""
profile gwenview @{exec_path} {
  include <abstractions/base-strict>
  # include <abstractions/deny-sensitive-home>
  include <abstractions/disks-read>
  include <abstractions/graphics>
  include <abstractions/kde-strict>
  include <abstractions/oneapi>
  # include <abstractions/private-files-strict>
  include <abstractions/thumbnails-cache-write>


  # set rlimit
  # userns
  # capability
  # network
  network netlink raw,
  # mount
  # remount
  # umount
  # pivot_root
  # change_profile
  # mqueue
  # signal
  signal send set=term peer=kioworker,
  # ptrace
  # unix
  # io_uring
  # dbus

  @{exec_path} mr,

  # @{bin}/, @{lib}/, /opt/
  @{lib}/kf6/kioworker Px,

  # /usr/share
  /usr/share/gwenview/** r,
  /usr/share/kImageAnnotator/translations/kImageAnnotator_de.qm r,
  /usr/share/thumbnailers/{,**} r,


  # /etc
  /etc/fstab r,
  /etc/machine-id r,

  # /, /var, /boot
  / r,

  #owner @{HOME}/[^.]**/ r,
  owner @{user_dirs}{,/[^.]**/} r,
  #owner @{HOME} r,
  #owner @{HOME}/[^.]/**/*.@{image_ext} rwl,
  owner @{user_dirs}/@{image_files} rwl,
  #deny owner @{user_dirs}/^@{image_files} rwl,
  #deny owner @{HOME}/*[^.@\{image_ext}] rwl,

  # @{user_cache_dirs}, @{user_config_dirs}, @{user_share_dirs}

  owner @{user_cache_dirs}/fontconfig/*-le64.cache-@{int} r,


  owner @{user_config_dirs}/#@{int8} rw,
  owner @{user_config_dirs}/gwenviewrc rw,
  owner @{user_config_dirs}/gwenviewrc.@{rand6} rwl,
  owner @{user_config_dirs}/gwenviewrc.lock rwk,
  owner @{user_config_dirs}/baloofileinformationrc r,

  owner @{user_share_dirs}/gwenview/recentfolders/{,**} rwkl,
  owner @{user_share_dirs}/baloo/index rw,
  owner @{user_share_dirs}/baloo/index-lock rwk,
  owner @{user_share_dirs}/kxmlgui5/gwenview/gwenviewui.rc r,
  owner @{user_share_dirs}/user-places.xbel r,

  owner @{user_state_dirs}/#@{int8} rw,
  owner @{user_state_dirs}/gwenviewstaterc rw,
  owner @{user_state_dirs}/gwenviewstaterc.@{rand6} rwl,
  owner @{user_state_dirs}/gwenviewstaterc.lock rwk,

  # /tmp/, @{run}/, /dev/shm/
  owner /tmp/pid-@{int}/ r,
  owner /tmp/pid-@{int}/*.@{image_ext} r,

  owner @{run}/user/@{uid}/#@{int} rw,
  owner @{run}/user/@{uid}/gwenview*.@{int}.kioworker.socket wl,
  owner @{run}/user/@{uid}/gvfs/ r,

  /dev/shm/ r,

  # @{sys}/
  @{sys}/bus/ r,
  @{sys}/bus/usb/devices/ r,

  # @{PROC}/
        @{PROC}/sys/kernel/random/boot_id  r,
  owner @{PROC}/@{pid}/fd/ r,
  owner @{PROC}/@{pid}/mountinfo r,


  # /dev/
  /dev/tty r,

  # deny
 

  include if exists <local/gwenview>
}

# vim:syntax=apparmor
```
FWIW, I've tried with ps and journalctl -f to find related processes that might explain this behavior but haven't detected anything meaningful.

valoq:
`@{lib}/kf6/kioworker Px,`

That would be my first guess

>  general-purpose worker for various protocols, such as accessing files on a local disk

valoq:
`owner @{user_dirs}{,/[^.]**/} r,`
This seems to also be an issue.

When I replace this with:
`owner @{HOME} r,`

It works as intended.

When the profile only includes these two lines of user files:

```
owner @{HOME} r,
owner @{user_dirs}/@{image_files} rwl,
```

I can select pdf files in $HOME for demonstration purposes (since listing files there is permitted by the first line) but get a permission denied when I try to open anything but image files.

curiosityseeker:
If I change `@{lib}/kf6/kioworker Px,` to `@{lib}/kf6/kioworker ix,` I can no longer open those odt files - good.

And if I use  `owner @{HOME} r,` I can no longer open pdf files - good. However, I can no longer open sub-folders in @{HOME}. So it has to be completed with `owner @{HOME}/[^.]**/ r,`. 

Thanks for helping to find a solution. This illustrates that finding a solution that cannot be bypassed is sometimes a bit tricky.

valoq:
`owner @{user_dirs}/[^.]**/ r,` 
Works as well, the problem was the line

`owner @{user_dirs}{,/[^.]**/} r,`
since it allowed empty expansion in the brackets, leading to effectively @HOME/**

curiosityseeker:
> `owner @{user_dirs}/[^.]**/ r,` Works as well, 

Not here. I cannot open @{HOME}

EDIT: This works if I add `owner @{user_dirs} r,`.

Unfortunately, with both rules I can open the odt files again :angry:

valoq:
Oh I added `owner @{HOME}/ r,` too

valoq:
> > `owner @{user_dirs}/[^.]**/ r,` Works as well,
> 
> Not here. I cannot open @{HOME}
> 
> EDIT: This works if I add `owner @{user_dirs} r,`.
> 
> Unfortunately, with both rules I can open the odt files again 😠

add the / at the end

curiosityseeker:
See my edit above. The only rules that work reliably for me are:

```
owner @{HOME} r,
owner @{HOME}/[^.]**/ r,
```

curiosityseeker:
> > > `owner @{user_dirs}/[^.]**/ r,` Works as well,
> > 
> > 
> > Not here. I cannot open @{HOME}
> > EDIT: This works if I add `owner @{user_dirs} r,`.
> > Unfortunately, with both rules I can open the odt files again 😠
> 
> add the / at the end

Ah - that did it! Thanks again!

apparmor.d/tunables/multiarch.d/user-dirs
```diff
@@ -0,0 +1,10 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2021-2025 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+# The user directories defined below permit access to user file paths. In combination with, e.g.,
+# the variables defined in .../tunables/multiarch.d/files they allow for rules like:
+#    owner @{user_dirs}/@{image_files} rw,
+
+@{user_dirs} =  @{HOME} @{HOME}/[^.]**
+@{user_dirs} += @{MOUNTS} @{MOUNTS}/[^.]**
```
cboltz: Please add a trailing `/` to the user_dirs definition so that it's limited to directories. Without that, it allows files and directories.

valoq: Yes, I was just experimenting with this and its a good idea to limit it to directories.

From what I tested just now it works as expected

valoq:
@user_dirs now only allows access to directories, not files.
To access files, either the @(type_files) is required or the profile needs to explicitly list 
`@{user_dirs}/[^.]**`

valoq:
This should be ready to merge now. @roddhjav let me know if any more changes are needed.
As soon as this is merged I will implement the new variables in existing profiles that use specific file types

roddhjav:
Let me think about it, feel free to ping me up in a week or two if I do not reply again.

I am not sure this solution would solve the problem in https://github.com/roddhjav/apparmor.d/discussions/873#discussioncomment-14709555.

We really need a solution that can move to really paranoid to really relaxed only by changing a variable. And we need to be able to easily apply it everywhere.

valoq:
> We really need a solution that can move to really paranoid to really relaxed only by changing a variable. And we need to be able to easily apply it everywhere.

Im not too sure we need that anymore with the file type filter.
I have gone ahead and implemented the new files variables where possible to better illustrate how it works.

The existing file categories can still be tightened in some areas to allow better control as well.

With the new types I feel this aligns quite well with paranoid mindset while also avoiding overblocking, so I don't see any need for multiple variants anymore.

Edit:
Actually this approach allows to change the access based on a single variable, through user_dirs
By setting this variable to specific directories, it will apply to all profiles.

apparmor.d/tunables/multiarch.d/user-dirs
```diff
@@ -0,0 +1,10 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2021-2025 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+# The user directories defined below permit access to user file paths. In combination with, e.g.,
+# the variables defined in .../tunables/multiarch.d/files they allow for rules like:
+#    owner @{user_dirs}/@{image_files} rw,
+
+@{user_dirs} =  @{HOME} @{HOME}/[^.]**/
+@{user_dirs} += @{MOUNTS} @{MOUNTS}/[^.]**/
```
roddhjav: You need the directory mark: `@{HOME}/`, `@{MOUNTS}/`

roddhjav:
@cboltz @gegarcia @jrjohansen Do you have any input on this?

I like the concept, but I would like to get some feedback from you.

valoq:
> Let me think about it, feel free to ping me up in a week or two if I do not reply again.

Ping

jrjohansen:
So let me start with long term direction for apparmor. and its policy  valoq's comment of getting us [closer to the android model](https://github.com/roddhjav/apparmor.d/pull/906#issuecomment-3415510983), is indeed the right direction.

Ideally we refactor policy into includes/interfaces that have an easy way to enable/disable on a per profile basis. So a user interface can present the user a check box, to enable/disable user-files, images, private files, ...

Ideally policy authors may ship with some includes/interfaces enabled by default but a user can over ride. Basically we need a system that allows tracking user changes, vs. default policy options.

Obviously this is going to mean using variables/includes, but also a few other things.

- An overlay for policy and includes: includes already allow searching multiple policy locations in order, this is going to be improved and extended to also profile files. This will allow a user to define a variable that will override the system variable, or have rules, or policies that over ride system profile files.

- Variables are going to pick up conditional assignment ?= and overriding assignment := to also help with some of the above

This will allow user to have to override default policy vars like say image extensions if they want, beyond providing simple gui check box support.

- Another thing that is coming is parametric include macros, so that any include can be reused with different variables. This will help with customization, as well, so we don't need to worry about user var defines being a global variable change.

- There is work towards runtime kernel variables, so that @{HOME}, and similar variables can have multiple instances, and respond dynamically. How this would work for a user is pam_apparmor (if configured for the user) will set-up @{HOME} as a kernel session variable for the user, and set its value based on the auth system (eg what is in /etc/passwd, or what ever auth system is in place).

so we should be trying to make sure we are using owner with @{HOME} or anything that is supposed to be a user-file access rule to help with the above transition.

Generally I prefer positive patterns for rule access instead of relying on deny to subtract access, because it makes extending policy with new rules so much easier. Rule priority added in 4.1 is a necessary crutch, but we should keep its use to a minimum. That is all a long winded way of saying, I like the pattern I see here to exclude private files. The other way to go would be to use tighter scoped boolean operations but those look like they won't land until 5.1 or maybe even 6.0

- As for prompt that was already mentioned. This too is going to come into play but it is going to take a while to get it upstream. The planned use here is again is to move closer to mac, and android. Applications should use portals where available or can be updated, but when they can't prompt will come into play.

Prompt will allow for multiple backends, not just snapd. And we do plan to start leveraging it. In addition to prompt there will be meta-data support so that a given prompt can carry addition information to userspace that it can choose to use. An example where it could come into play here would be the image files, and file extensions. Access to an a file with an image extension could trigger a prompt, where a backend daemon could check that it is actually an image file (run file on the file),  or similar actions before actually granting access.

I have only done a first quick pass on the MR, and I think it is moving in the general direction of above. I need to spend some more time with it. But generally I like what it is doing.

valoq:
Gentle ping for review

valoq:
I have been using this change on a daily basis for three month now without issues. It should be ok to be merged

roddhjav:
Don't worry, it is going to be merged. I am now very optimistic about this change. Howerver; please allow time. I don't want to merge this change that fully change the way files is allowed to be accessed (i.e. that add possible breaking changes, require lot of documentations and tests) while also breaking changes on other part of the project (cf #986 #464)

curiosityseeker:
While I love this approach, there is one drawback: For every file **not** matching the `@{user_dirs}/@{*_files}` rule in the respective applications the logs will be spammed with numerous denials. 

I don't know how to avoid this. Even if a rule like 
```
deny  [^@{user_dirs}/@{document_files} ] r, 
```
would work (it doesn't because of a syntax error) we couldn't use it as it would also block access to `@{user_config_dirs}` etc.

valoq:
> While I love this approach, there is one drawback: For every file **not** matching the `@{user_dirs}/@{*_files}` rule in the respective applications the logs will be spammed with numerous denials.

What do you mean? Most profiles in this PR I use daily and they dont leave any deny enties in the logs. The new variables are meant to cover all common types that are required. If a profile requires additional rare file types not specified in e.g. `document_files`, they can be added like `@{user_dirs}/[^.]*.[pP][dD][fF] r,`

curiosityseeker:
> What do you mean? Most profiles in this PR I use daily and they dont leave any deny enties in the logs.

It seems I didn't express myself clearly. Let's say you're using an image reader with `owner @{user_dirs}/@{image_files} rwl,` and the same user dirs contain other file types like, e.g., documents, or files without extension. Then you will get a lot of read denials in the log.

This is nothing dramatic but perhaps a bit annoying/confusing particularly for newbies. This could, theoretically, be overcome by an appropriate deny rule ([as operations that trigger "deny" rules are not logged by default](https://man.archlinux.org/man/extra/apparmor/apparmor.7.en#Turn_off_deny_audit_quieting)). But I haven't found such a working deny rule.

valoq:
I see what you mean. Based on some quick tests, it seems this issue depends on the application though. Im not sure if there is a valid reason but I actually tend towards arguing this is a bug in the affected applications since they try to get information about files they have no biusiness with. That does not change the problem of deny logs for us though.

Opening an image with `imv` in $HOME will lead to denied file access logs regarding files such as `.bash_history`which seems rather strange. Why would imv try to access the file? The general infomation about the files in the directory are already permidded through `@{user_dirs} r,` I think. Maybe it actually runs a mime file check on all files to determine if it should parse them as image files. Probably something we have to deal with then.

@roddhjav Any ideas about how to avoid this making it into the logs?

cboltz:
> [...] this is a bug in the affected applications since they try to get information about files they have no biusiness with. 

Not sure if this is really the case here, but - sounds like what all the KDE file dialogs do ;-)

> Opening an image with `imv` in $HOME will lead to denied file access logs regarding files such as `.bash_history`which seems rather strange. [...]
> 
> @roddhjav Any ideas about how to avoid this making it into the logs?

To deny access to the dot files, `include <abstractions/private-files-strict>`

You could also use a deny rule with low priority so that the allow rules can "win":

    priority=-1 deny @{user_dirs}/** r,

Note that this needs at least AppArmor 4.1.

Obviously silencing the logs with such a deny rule means you won't get log entries for the denied permissions anymore - which can make finding problems with the profile quite hard.

curiosityseeker:
> You could also use a deny rule with low priority so that the allow rules can "win":
> 
> ```
> priority=-1 deny @{user_dirs}/** r,
> ```

Cool :cool: I wasn't really aware of this (although @roddhjav had mentioned it somewhere recently). I've tested it with my `gwenview` profile and it works as advertised :smiley:  Thanks for this hint - this seems to be the solution.

> Obviously silencing the logs with such a deny rule means you won't get log entries for the denied permissions anymore - which can make finding problems with the profile quite hard.

Yes, but for debugging purposes this can be temporarily commented.

curiosityseeker:
Another drawback are files without extension. Text files are often saved without an extension, hence an application using `@{user_dirs}/@{text_files} rwl,` would not be able to read it. Is there a solution for such cases?

valoq:
> Another drawback are files without extension. Text files are often saved without an extension, hence an application using `@{user_dirs}/@{text_files} rwl,` would not be able to read it. Is there a solution for such cases?

Not really. Text files are a special case though, all other file types are pretty much always with an extension and that is also the focus of the MR. Text editors usually dont have any relevant attack surface anyway, so allowing all files should be fine in most cases. 


Regarding the supressed logs, we could add a special log entry whenever a process runs that uses a profile with surpressed logs. That would indicate somethign to consider when debugging the profile.

jrjohansen:
>> Obviously silencing the logs with such a deny rule means you won't get log entries for the denied permissions anymore - which can make finding problems with the profile quite hard.
>
> Yes, but for debugging purposes this can be temporarily commented.

for debugging purposes you can also use

```
sudo bash -c "echo -n noquiet > /sys/module/apparmor/parameters/audit"
```

which will turn off deny quieting

jrjohansen:
some future considerations (all of this is some phase of dev), that are at least tangentally related. This doesn't really affect the current PR, but can be something to consider when making design decisions.

[pam_apparmor is picking up abilities to set/modify the @{HOME} variable](https://gitlab.com/apparmor/apparmor/-/merge_requests/1917). This will help with non-standard $HOME locations, and dynamic kerberos/ldap/... setups
- first step: is the global abstraction. This can work on all systems but isn't per user.
- second step is a per user abstraction. This will require the user conditional (future), and identity delegation (experimental in 5.0)
- third step is per user kernel variables (future)

Dealing with files and file types. There are several extensions coming that will help.
- bool operations on rules:  rule priority can work for this but is brittle and relies on rule overlap for rules that may be far away, or hidden in different includes etc.  boolean operations will let you be more precise a the rule level. ```rw /** except /**.txt``` etc
- extended RE matching: again ore flexibility in what a rule can specify
- prompt: send the question about access to the user. Prompt also opens the possibility of a userspace daemon running a type check on the file before granting access
- better/more file conditionals: device type, extended ownership, inode, matching and setting a security label
- user defined policy: allow a user to define policy for their applications. This can be anything from setting up their own profiles that will be stacked against the system profile, to allowing users to define bounded (the system gets to define the bounds) extension to system profiles.  Such user extended profiles become variants of the system profile, and each user can have a unique variant.

curiosityseeker:
@jrjohansen Intriguing news- can‘t wait for these improvements 😉 Thanks for sharing.

valoq:
@roddhjav Let me know what the plans for this PR currently are and I can pick it up to get this merged

Issue #903, opened 2025-10-11
Alternative glycin integration


valoq:
The recent changes of glycin and the gdk-pixbuf library required the addition of extensive permissions to many profiles to allow the library internal bwrap sandbox to run. However the glycin code actually includes a [fallback](https://gitlab.gnome.org/GNOME/glycin/-/blob/main/glycin/src/sandbox.rs?ref_type=heads#L676) to run without bwrap when it is not available.

When the glycin-loader was not available in gtk applications it caused crashes even when the alternative execution path without bwrap was executed.
However, when allowing the glycin-loader permissions (like filesystem access) while blocking bwrap, applications like firefox run just fine.

This gives us an alternative approach to avoid stacked permissions for bwrap:

- adding a glycin abstraction without bwrap to the gtk abstraction allows any application using gtk to run without issues and without gaining problematic permissions by using the alternative execution path - The applications may not include permissions for @{bin}/bwrap to avoid crashing during the check

- Applications without the nnp flag set can still transition to dedicated glycin profiles without stacking the bwrap permissions

roddhjav:
It is a good point. So it seems that denying the execution of `/usr/bin/true` should disable the use of the sandbox. It would indeed solve the issue in app with nnp flag enabled.

Meanwhile, it is a bit sad to not be able to have both. See https://github.com/roddhjav/apparmor.d/issues/881#issuecomment-3360066162

valoq:
> So it seems that denying the execution of `/usr/bin/true` should disable the use of the sandbox

Actually the true command is attempted inside the bwrap container. To avoid any crashes, bwrap itself has to be blocked from execution in the first place. It should be easy to integrate by adding a glycin-pixbuf abstraction to the GTK abstraction without bwrap execution permission included. Applications that want to use the full bwrap feature, can include a transition to glycin which includes the previous abstraction and adds bwrap execution permissions.

> Meanwhile, it is a bit sad to not be able to have both. See https://github.com/roddhjav/apparmor.d/issues/881#issuecomment-3360066162

I agree with the statement, however at the moment we are talking about the use of glycin in gdk-pixbuf which uses it to render icons. Perhaps someone can point out if I missed something but there does not appear to be any realistic threat model behind this, or in other words, there is no need for a sandboxed renderer for icons that are installed with the application itself. 
[...]

PR #918, opened 2025-10-26
add glycin-fallback

valoq:
Addresses #903 

Profiles with the nnp flag such as firefox can replace the line
`include <abstractions/app/bwrap-glycin>`
with
`include <abstractions/app/glycin>`

which will allow glycin to run without bwrap

### DIFF
--- a/apparmor.d/abstractions/common/electron
+++ b/apparmor.d/abstractions/common/electron
@@ -20,6 +20,7 @@
 
   abi <abi/4.0>,
 
+  include <abstractions/app/glycin>
   include <abstractions/bus-session>
   include <abstractions/common/chromium>
   include <abstractions/dconf-write>


roddhjav:
How it this changing anything? Shouldn't you deny execution of bwrap instead?

Anyway, I think it is a good solution for now. We may later use apparmor namespace for an even better integration (if its works).

valoq:
From what I tested, these permissions are necessary to avoid crashes but the execution path in glycin still falls back to the non bwrap function if the permissions in `abstractions/bwrap` are not available

I think we can add `<abstractions/app/glycin>` to the gtk abstraction in order to avoid more bug reports from gtk applications since they will eventually all use the new gdk-pixbuf version. If new uses of glycin appear we may need to review this however to allow the transition to the separate bwrap-glycin profile. Or perhaps we could add this already by including the transition to the dedicated glycin profile which fails for nnp applications but succeeds for everyone else? If that works it may be the best option with the least future maintenance while also allowing glycins internal sandbox where possible.

It should also be mentioned that the current solution still triggers log complains about the missing bwrap execution permissions like userns, which cannot be silenced by deny since that would block userns in parent processes like firefox. If there is another way to silence those complains, let me know.

valoq:
I added the glycin permissions (without the bwrap requirements) to the gtk-strict abstraction to automatically include it where needed. This should address the issue for all affected applications.

roddhjav:
Sorry for the late response, they have been some changes on this matter recently.
Your solution works, but I have a better one that fix (almost) all issues.

More than the situation caused by nnp set in parent apps, the most important thing to understand is that `bwrap` raise a few highly privileged capabilities. And these are the problem. However, **no such caps are actually used...**

`bwrap` creates a user namespace. In this namespace, it needs these caps to mount things and setup networking. The caps are not needed in the parent namespace (i.e. the real root) and only used by the *fake* root in this namespace. The problem is that **apparmor does not see it** and interprets the access as if they were coming from the real root.

But it can understand it. By setting an [**apparmor namespace**](
https://gitlab.com/apparmor/apparmor/-/wikis/AppArmorNamespaces). The idea is to set an apparmor namespace to confine everything that is in the user namespace. This will have the effect of completely bypass the nnp flag, as the app is not actually raising anything it does not already have.

From a security point of view, it is really good as:
- `firefox` (or any sandboxed app that use glycin loader) will transition to the `glycin` namespace upon the launch of bwrap.
- `:glycin:bwrap` confine the user namespace and sandbox creation for glycin, limiting vulnerabilities found in user namespace.
- `:glycin:loaders` confine the already sandboxed loaders. The loader is actually confined by `:glycin:bwrap//&loaders`, as nnp still applies in the ns when it is set by `:glycin:bwrap`. No problem.
- For non sandboxed app, it does not change that much, the loader is now confined by `:glycin:bwrap//&loaders`, instead of  `&glycin//loaders`.

We still have to deal with the fact that bwrap can be used to sandbox other programs than glycin. It is currently not a big issue as we can handle the few cases where it applies (gnome-software, flatpak, the thumbnailers). We may need apparmor integration in bwrap for this. But that is another subject for another time (year).

The only downside, is that it will have to wait for the next patch release of apparmor to be integrated, as while working on this I found a (now fixed) bug in apparmor (see https://gitlab.com/apparmor/apparmor/-/merge_requests/1829)

valoq:
If I understand this correctly, upstream has backported the patches required for this into the 4.1 branch and we can use this solution starting with the next minor release of apparmor 4.1 ?

roddhjav:
Yes, it is the plan.

PR #920, opened 2025-10-26
Changes for mpv profile

no-madsoul:
I'm running Debian system with Radeon graphics. Recently after update to Debian Trixie I've tried to roll some apparmor profiles from project's repo in complain mode and  got some complaining in auditd log that mpv process was trying to get access to radv mesa cache dir, udmabuf device and to script folder which is one level below /etc/mpv.
So this pull request adds necessary changes to mesa.d/complete abstraction and  mpv profile.


roddhjav:
Thanks, merged!

PR #928, opened 2025-11-03
hwmon abstraction consistency for profiles

no-madsoul:
Hi!
   I've started playing around with waybar profile and noticed that waybar tries to read temperature measurements from /sys/devices/.../hwmon(0..9). 
   I've included abstractions/hwmon to waybar profile and checked how it's organized in other profiles with "hwmon" in text...
   Well, i've ended up with this pull request where  i've tried to generalize rules in abstractions/hwmon and change different links to /sys/devices/.../hwmon in some profiles for "include <abstractions/hwmon>".
 
Best regards,
Dmitry.

### DIFF
--- a/apparmor.d/abstractions/hwmon
+++ b/apparmor.d/abstractions/hwmon
@@ -5,11 +5,37 @@
   abi <abi/4.0>,
 
   @{sys}/class/hwmon/ r,
+  @{sys}/devices/**/hwmon/ r,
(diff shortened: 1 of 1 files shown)


no-madsoul:
Sorry for bothering, but after fresh look at pull request I've realized that it should be split in parts by one profile in each pull request, besides i should fix trailing white spaces and bad indents. Me bad :-(

PR #929, opened 2025-11-04
Hwmon abstraction change proposition

no-madsoul:
In attempt to unify profiles that grant permissions to work with sys/hwmon nodes I added hwmon-full abstraction with full specs and permissions in accordance with https://www.kernel.org/doc/Documentation/hwmon/sysfs-interface. Previous hwmon abstraction was recreated from hwmon-full with only "read" permissions.


roddhjav:
Nice work, thanks. The initial `hwmon` abs needed more work, it is exactly this.

A few comments:
- The full version should include `hwmon` so deduplicated rules can be removed.
-  `{,hwmon/}` is not required. Simplify it to `@{sys}/devices/**/hwmon@{int}/`
- That could be a tasks for me later, as most profiles do no need to monitor all hardware, it could be nice to have individual abs such as `hwmon-voltages`, `hwmon-fans`...
- In abstractions, do not hesitate to be verbose. E.g. a `@{sys}/devices/**/hwmon@{int}/temp@{int}_{emergency,emergency_hyst} r,` would be easier to read/maintain as:
```
  @{sys}/devices/**/hwmon@{int}/temp@{int}_emergency r,
  @{sys}/devices/**/hwmon@{int}/temp@{int}_emergency_hyst r,
```

no-madsoul:
Hi! Thank you.
As for verbosity of rules and removing {,hwmon/} points are taken and work is done. But before updating pull request I'd like to ask about rules de-duplication mentioned. As I was thinking: hwmon nodes according to docs has 3 states: read only, write only, read-write. So I decided include in hwmon read only and read-write with write permission removed. What was proposition for rules? Was it that hwmon has read only and read permissions and hwmon-full  has stacking(?) rules for corresponding hwmon with added write permission and write only rules and include <abstractions/hwmon>? Or better make something, for example, hwmon-read (read only, and read), hwmon-write and hwmon-full(includes from both previous)?

roddhjav:
Thanks a lot, merged.

All important information is here, I may split the main abstraction later, but I still need to decide on how to split it. For now, it looks very good.

PR #949, opened 2025-11-28
Update fwupd

JND94:
I was unable to update UEFI db
`DENIED  fwupd open @{sys}/firmware@{efi}/efivars/db-@{uuid} comm=fwupd requested_mask=a denied_mask=a`

I'm not sure if `@{efi}` var should be used here, also I don't know if it will work fine with append permission only, or just gran write since I personally just added `rw` to make it work


roddhjav:
Thanks, good catch

PR #951, opened 2025-12-04
fix(profile): plasma: browser integration artwork

lina-bh:
The `plasma-browser-integration` native messaging host writes the current browser's stream's artwork to a tmpfile (`O_TMPFILE` by way of `QTemporaryFile`) which is then read by `plasmashell`. At present in `main` neither ends of the communication succeed. These rules grant read/write/link on the tmpfile path format to `plasma-browser-integration-host` and read on it to `plasmashell`.


roddhjav:
Thanks, merged.

PR #953, opened 2025-12-06
Fix xdg_password_store variable inconsistency

Stoppedpuma:
Previously, it was a mixture of XDG_PASSWORDSTORE_DIR and XDG_PASSWORD_STORE_DIR which was a bit confusing as the documentation is currently wrong and would fail to enable Apparmor. this PR fixes that by changing all use to XDG_PASSWORD_STORE_DIR which matches the most common use of the format.

I can also flip this around and use XDG_PASSWORDSTORE_DIR if that's desired. I went with xdg_password_store_dir as it's the most common use both in apparmor.d, and used other places like the arch wiki.

### DIFF
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -42,7 +42,7 @@
 # Define user personal keyrings
 @{XDG_GPG_DIR}=".gnupg"
 @{XDG_SSH_DIR}=".ssh"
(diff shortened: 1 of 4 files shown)


roddhjav:
Good catch. However, you should fix it the other way around: fix the doc and keep the variables in the profiles as they are. It will avoid breaking compatibility.

Stoppedpuma:
Both of the password variables should now be consistent.

roddhjav:
Nice, thanks, merged!

PR #955, opened 2025-12-09
KDE on Debian 13

nobody43:


nobody43:
`error useless(apparmor.d/profiles-s-z/unix-chkpwd:30): rule already included in the base abstraction, remove it`
Can't see it. Nevertheless, it still throws denials with rule absent.

`error tunables(apparmor.d/profiles-g-l/goxray_cli:8): variable '@{arch}' must be used instead of: amd64`
I don't think so: attachment path is the most important part of the profile - it shouldn't be obscured too much. Change if you think otherwise.

`error header(apparmor.d/profiles-g-l/goxray_cli): missing header: '# Copyright (C) '`
Can't fix it for ethical reasons, you have to do it yourself.

apparmor.d/groups/kde/dolphin
```diff
@@ -77,26 +78,40 @@ profile dolphin @{exec_path} {
...
 
   owner @{user_cache_dirs}/ksplash/qmlcache/*.qmlc rwl -> @{user_cache_dirs}/dolphin/qmlcache/#@{int},
   owner @{user_cache_dirs}/ksplash/qmlcache/*.qmlc.@{rand6} rwl -> @{user_cache_dirs}/dolphin/qmlcache/#@{int},
 
   owner @{user_state_dirs}/dolphinstaterc{,.*} rwlk -> @{user_state_dirs}/#@{int},
 
-  owner @{tmp}/dolphin.@{rand6}{,.lock} rwlk,
+  owner @{tmp}/#@{int} rwk,
+  owner @{tmp}/dolphin.@{rand6} rwl -> @{tmp}/#@{int},
+  owner @{tmp}/dolphin.@{rand6}.@{rand6} l -> @{tmp}/#@{int},
```
nobody43, line 112: That's what I saw in the dmesg. Maybe a dolphins bug.

apparmor.d/profiles-s-z/telegram-desktop
```diff
@@ -7,11 +7,12 @@ abi <abi/4.0>,
 
 include <tunables/global>
 
-@{exec_path} = @{bin}/telegram-desktop @{bin}/Telegram
+@{exec_path} = @{bin}/telegram-desktop @{bin}/Telegram /opt/Telegram/Telegram
 profile telegram-desktop @{exec_path} {
-  include <abstractions/base>
+  include <abstractions/base-strict>
   include <abstractions/audio-client>
-  include <abstractions/consoles>
```
nobody43: This application and it's source is too shady, works perfectly fine without consoles access.

apparmor.d/groups/kde/dolphin
```diff
@@ -77,26 +78,40 @@ profile dolphin @{exec_path} {
...
 
   owner @{user_config_dirs}/#@{int} rw,
   owner @{user_config_dirs}/dolphinrc rwl -> @{user_config_dirs}/#@{int},
   owner @{user_config_dirs}/dolphinrc.@{rand6} rwl -> @{user_config_dirs}/#@{int},
   owner @{user_config_dirs}/dolphinrc.lock rwk,
-  owner @{user_config_dirs}/kde.org/#@{int} rw,
+  owner @{user_config_dirs}/kde.org/#@{int} rwk,
   owner @{user_config_dirs}/kde.org/UserFeedback.org.kde.dolphin.conf.lock rwk,
-  owner @{user_config_dirs}/kde.org/UserFeedback.org.kde.dolphin.conf{,.*} rwlk -> @{user_config_dirs}/kde.org/#@{int},
+  owner @{user_config_dirs}/kde.org/UserFeedback.org.kde.dolphin.conf.@{rand6} rwl -> @{user_config_dirs}/kde.org/#@{int},
```
nobody43: I haven't seen any calls to the base config, so I didn't made assumptions. I'll add the rules once I see them.

apparmor.d/profiles-s-z/telegram-desktop
roddhjav: Please do not use the base-strict abstraction directly. It gets selected at build time.

apparmor.d/groups/kde/dolphin
roddhjav: This won't work, I can assure you they are all needed.`

apparmor.d/profiles-s-z/xray
```diff
@@ -10,11 +10,15 @@ include <tunables/global>
 profile xray @{exec_path} flags=(attach_disconnected) {
   include <abstractions/base>
   include <abstractions/consoles>
+  include <abstractions/ssl_certs>
+  include <abstractions/openssl>
```
roddhjav: `abstractions/openssl` is included in the base abstraction already.

roddhjav:
Sorry for the late review. In general, it is good, I have however a few comments:

1. I am not adding empty profiles (that most likely will break things).  By the name of those, it is obvious they require more rules: 
   - rustdesk_polkit
   - rustdesk_startwm 

2. Please do **not** use `abstractions/base-strict` hardcoded in here. It is replaced by default during the build phase.

3. You may want to take an extra care about the number of named transition (`->`) you add in profile. We are limited by those (12 unique by profile), adding them to every link raise risk of compilation issue.

> error useless(apparmor.d/profiles-s-z/unix-chkpwd:30): rule already included in the base abstraction, remove it
> Can't see it. Nevertheless, it still throws denials with rule absent.

It is in: `abstractions/glibc` (itself in `base-strict`).

> error tunables(apparmor.d/profiles-g-l/goxray_cli:8): variable '@{arch}' must be used instead of: amd64
> I don't think so: attachment path is the most important part of the profile - it shouldn't be obscured too much. Change if you think otherwise.

No, it is a maintenance nightmare with the direct consequence of beaking app, which is far, far worse.

> error header(apparmor.d/profiles-g-l/goxray_cli): missing header: '# Copyright (C) '
> Can't fix it for ethical reasons, you have to do it yourself.

Fine with me. But you know that the "Copyright" only means it is from you (a thing we cannot deny: it will still be your work anyway), the licence is still GPL2.

apparmor.d/profiles-s-z/xray
nobody43: Can't see on my freshly installed Debian 13. Please recheck the compilation result on this system.

nobody43:
1. I'm using Rustdesk daily and can't see any rules which could be added there. Rustdesk is an example of notoriously badly written software.
2. Fixed with 1fd353a
3. I thought that only applies to profile transitions, not link access. I always test the .deb package build for such broad PRs

apparmor.d/groups/kde/dolphin
```diff
@@ -77,26 +78,39 @@ profile dolphin @{exec_path} {
...
-  owner @{user_config_dirs}/kde.org/UserFeedback.org.kde.dolphin.conf{,.*} rwlk -> @{user_config_dirs}/kde.org/#@{int},
+  owner @{user_config_dirs}/kdeglobals.@{rand6} l -> @{user_config_dirs}/#@{int},
+  owner @{user_config_dirs}/kdeglobals.lock k,
+  owner @{user_config_dirs}/kiorc l -> @{user_config_dirs}/#@{int},
+  owner @{user_config_dirs}/kiorc.lock k,
+  owner @{user_config_dirs}/kservicemenurc l -> @{user_config_dirs}/#@{int},
+  owner @{user_config_dirs}/kservicemenurc.lock k,
+  owner @{user_config_dirs}/ktrashrc l -> @{user_config_dirs}/#@{int},
+  owner @{user_config_dirs}/ktrashrc.lock k,
   owner @{user_config_dirs}/knfsshare.{,.@{rand6}} rwk,
```
nobody43, line 100: Even this? Looks very strange.

apparmor.d/profiles-s-z/telegram-desktop
nobody43: Fixed with 1fd353a27cfba64e92066695b7ab72b51caa623a

PR #1002, opened 2026-01-20
Vesktop Fixes

K-Hobert:
Camera access included, user's picture/video folders allowed RW and audio source discovery for screen sharing fixed


roddhjav:
Thanks, merged!

PR #1017, opened 2026-02-07
discord: allow autostart .desktop creation

JND94:
Discord app can enable autostart in "Linux settings"


roddhjav:
Thanks, merged!

PR #1018, opened 2026-02-07
Mullvad

curiosityseeker:
Mullvad VPN is now in the official Arch repos, hence some paths in the `mullvad-gui` and `mullvad-setup` profiles have to be added.

### DIFF
--- a/apparmor.d/groups/network/mullvad-gui
+++ b/apparmor.d/groups/network/mullvad-gui
@@ -26,7 +26,6 @@ profile mullvad-gui @{exec_path} flags=(attach_disconnected) {
   network netlink raw,
 
   @{exec_path} mrix,
-  
   @{bin}/gsettings rPx,
 
   @{open_path} rPx -> child-open-browsers,


roddhjav:
Nice, thank. Can you fix the trailing whitespace so I can merge it.

curiosityseeker:
You refer to the ones after @{open_path} ? Done!

roddhjav:
No, that does not matter. The issue is the trailing whitespace line 29:
```
error trailing(apparmor.d/groups/network/mullvad-gui:29): line has trailing whitespace
```

apparmor.d/groups/network/mullvad-gui
```diff
@@ -26,8 +26,10 @@ profile mullvad-gui @{exec_path} flags=(attach_disconnected) {
   network netlink raw,
 
   @{exec_path} mrix,
+
```
roddhjav:
```suggestion

```

curiosityseeker:
`error trailing(apparmor.d/groups/network/mullvad-gui:29): line has trailing whitespace`

Silly question: How do you get such a notification? Do you use a script for that, and can we as contributors use that as well?

roddhjav:
For now, this it this (basic) check script: https://github.com/roddhjav/apparmor.d/blob/main/tests/check.sh. It runs in the Github action jobs (it is why the job failed). Later they will be a proper linter (but I need a full AST of the profile for this).

You can use it with `just check`, and not PR will be merged with errors raised from this script. So yes, you can use it.

curiosityseeker:
Cool - thanks! Just noticed: Shouldn't line 179 rather be

`["bash"]="shells"`

?

roddhjav:
You are perfectly right. Thanks for catching this.

PR #1023, opened 2026-02-11
cups-backend-usb profile

skarpinis:
Minor adjustment to allow CUPS to read USB devices.


roddhjav:
This is already included in `abstractions/devices-usb`

skarpinis:
> This is already included in `abstractions/devices-usb`

I see just noticed sorry about that!

PR #1024, opened 2026-02-11
Minor tweak

skarpinis:
added allow read access for nvidia.


roddhjav:
Good tweak thanks, however, I don't see anything regarding nvidia.

skarpinis:
> Good tweak thanks, however, I don't see anything regarding nvidia.

Thank you! The file is located at /etc/cdi/nvidia.yaml. :)

PR #1025, opened 2026-02-12
small update to some profiles.

skarpinis:
Fix for polkit-agent-helper to allow authentication, along with a small adjustment for the Thunar file manager to read drivers.


roddhjav:
Thanks, merged. Please for next PR, use the lasted main branch as base.

PR #1031, opened 2026-02-15
Fix check.sh

curiosityseeker:
As mentioned in https://github.com/roddhjav/apparmor.d/pull/1018#issuecomment-3884698006


roddhjav:
Nice, thanks. Merged!

PR #1032, opened 2026-02-15
minor update to various profiles

skarpinis:
Few minor tweaks and fix for docker with cuda.

### DIFF
--- a/apparmor.d/abstractions/app/chromium
+++ b/apparmor.d/abstractions/app/chromium
@@ -121,7 +121,8 @@
   owner @{config_dirs}/** rwk,
   owner @{config_dirs}/WidevineCdm/*/_platform_specific/linux_*/libwidevinecdm.so mrw,
 
-  owner @{cache_dirs}/{,**} rw,
+  owner @{cache_dirs}/ rw,
+  owner @{cache_dirs}/** rwk,
 
   owner @{user_config_dirs}/kioslaverc r,
   owner @{user_config_dirs}/menus/applications-merged/ r,
@@ -139,7 +140,7 @@
 
         /tmp/ r,
         /var/tmp/ r,
-  owner @{tmp}/@{domain}.*/{,**} rw,
+  owner @{tmp}/{,.}@{domain}.*/{,**} rw,
   owner @{tmp}/@{name}-crashlog-@{int}-@{int}.txt rw,
   owner @{tmp}/cache/Default/ rw,
   owner @{tmp}/cache/Default/** rwk,


apparmor.d/abstractions/app/chromium
```diff
@@ -112,6 +112,7 @@
   owner @{HOME}/ r,
 
   owner @{user_cache_dirs}/gtk-3.0/**/*.cache r,
+  owner @{user_cache_dirs}/google-chrome/Default/image_cache/cached_image_fetcher_images/LOCK k,
```
JND94, line 115: This seems to be Google Chrome specific and I don't think it belongs to chromium abstraction

apparmor.d/abstractions/app/chromium
```diff
@@ -138,7 +139,7 @@
 
         /tmp/ r,
         /var/tmp/ r,
-  owner @{tmp}/.@{domain}.*/{,**} rw,
```
JND94, line 141: Chromium can create temp folder that starts with `.@{domain}`

apparmor.d/abstractions/app/chromium
skarpinis: I am not entirely certain, but I believe they belong to the Chromium family. They all appear to include that abstraction. Do we need to create a Google Chrome profile separately?

apparmor.d/abstractions/app/chromium
skarpinis: It’s not sufficient: DENIED - brave mkdir owner /tmp/org.chromium.Chromium.BraveComponentUpdater_chrome_url_fetcher_.8MicTU/ comm=ThreadPoolForeground requested_mask=c denied_mask=c. Does this indicate that it is not part of Chromium again?

roddhjav: I made similar changes to the `common/chromium` abs (abs used by `app/chromium`). Both need to be supported:
```
owner @{tmp}/{,.}@{domain}.*/{,**} rw,
```

apparmor.d/abstractions/app/chromium
skarpinis: @roddhjav Do you want I will move this to the chrome profile?

roddhjav: It should be in the chromium abs; but not hardcoded this way. In the chromium abs, this path is handled by `cache_dirs`:

https://github.com/roddhjav/apparmor.d/blob/d8e6e4966cc7a2ae0d936ee728cddfee06adb321/apparmor.d/abstractions/app/chromium#L123

Should become:
```
  owner @{cache_dirs}/ rw,
  owner @{cache_dirs}/** rwk,
```

roddhjav:
Thanks for this PR.

Regarding cuda, can you provide example of setup that raised this so I can test it on my side. This cannot be merged as it as `@{bin}/nvidia-container-runtime ix,` should most likelly transition to another profile (and thus, the added rules should not be in this profile, but in the transitionned one). Anyway, as this one is not a small change, it will have to be tested and merged in another PR.

Issue #1033, opened 2026-02-18
KDE Plasma “Picture of the Day” / other HTTPS KIO fetches fail due to kioworker AppArmor rule denying /etc/** reads


ehbstudentadam:
On KDE Plasma (CachyOS/Arch, KF6), Plasma Addons that fetch remote content via KIO (e.g. Wallpaper → Picture of the Day providers like APOD/Bing) fail for all providers. Plasma logs show 0-byte downloads, e.g.:

- `plasmashell: ... job URL = QUrl("https://apod.nasa.gov/apod/") data size = 0`
- `"apod" ... failed to fetch the remote wallpaper. Please check your Internet connection or system date.`

Environment

- KDE Plasma 6 / KF6
- KIO HTTP worker is /usr/lib/kf6/kioworker using kio_http.so (https)
- AppArmor enabled, kioworker profile enforced

Audit log shows kioworker gets EACCES trying to open world-readable TLS-related files, which breaks HTTPS:
- [log.txt](https://github.com/user-attachments/files/25397865/log.txt)

### Root cause

In /etc/apparmor.d/kioworker, the profile includes abstractions that allow TLS/name service:

```
include <abstractions/nameservice-strict>
include <abstractions/ssl_certs>
```
…but later it applies a broad deny:

```
deny /etc/{,**} r,
```
Because AppArmor deny rules override allows, this blanket deny blocks the SSL cert store and OpenSSL config under /etc/ssl/**, causing HTTPS KIO fetches (POTD providers, etc.) to fail.

### Fix

Change the “silence non user’s data” rule to deny writes to /etc, not reads:

```
--- a/apparmor.d/kioworker
+++ b/apparmor.d/kioworker
@@ -... +... @@
-  deny /etc/{,**} r,
+  deny /etc/{,**} w,
```

After this change, remote wallpaper providers immediately start working again.

skarpinis:
Thanks, it should be fixed now. #1039

PR #1039, opened 2026-02-21
Fix for #1033

skarpinis:
Fixed KIO HTTPS failures in kioworker AppArmor profile by changing the blanket /etc deny from read to write. The profile previously blocked reads under /etc (including /etc/ssl and OpenSSL configs) which caused 0-byte downloads; updating
deny /etc/{,} r,
to
deny /etc/{,} w,
preserves required reads (certs/config) while still preventing writes, restoring remote-content (APOD/Bing POTD) downloads.

### DIFF
--- a/apparmor.d/groups/kde/kioworker
+++ b/apparmor.d/groups/kde/kioworker
@@ -75,7 +75,7 @@ profile kioworker @{exec_path} {
 
   # Silence non user's data
   deny @{efi}/{,**} r,
-  deny /etc/{,**} r,
+  deny /etc/{,**} w,
   deny /opt/{,**} r,
   deny /root/{,**} r,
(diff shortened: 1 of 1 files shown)


JND94:
does even kioworker try to write in /etc ? that sounds unlikely as it's spawned as user process

roddhjav:
kioworker is used file filebrowser to query/write file. See https://apparmor.pujol.io/configuration/#file-browsers

JND94:
in the same way that there's no need to specify a write deny to /usr or any other root owned folder... kioworker doesn't try to write on /etc why should logs be silenced with a specific deny

roddhjav:
That was mostly to cleanup logs in complain mode. Not strictly required in enforce mode.

PR #1041, opened 2026-02-24
New PR based from #1032

skarpinis:
I have removed Docker with CUDA. Until you test it and improve the implementation, I will continue to use local rules.


roddhjav:
Thanks, merged!

PR #1045, opened 2026-02-27
ssh-keyscan: initial profile

doublez13:
Add an AppArmor profile for ssh-keyscan.

ssh-keyscan has the ability to read a list of IPs, hostnames, and networks from a file. I wasn't sure how restrictive the file location should be. That's what the `owner /{,**} r,`  rule is for. Other than taking input from a file, no other direct filesystem access is needed.

For example:
```
echo 'ssh.example.com' > /tmp/hosts
ssh-keyscan -f /tmp/hosts
```

Let me know if this should be restricted.


roddhjav:
Thanks for this PR, however, there is already a profile for this program upstream: https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor.d/ssh-keyscan

doublez13:
Lol ah jeez I haven't refreshed against the upstream profiles in awhile now. Didn't realize it had been over a year now.

PR #1047, opened 2026-02-28
fix dockerd & docker-proxy on debian

skarpinis:
Fixed the AppArmor profile for dockerd and the docker-proxy path on Debian; adjusted the run-parts profile to allow ClamAV.

### DIFF
--- a/apparmor.d/groups/virt/docker-proxy
+++ b/apparmor.d/groups/virt/docker-proxy
@@ -9,6 +9,7 @@ include <tunables/global>
 @{exec_path} = @{bin}/docker-proxy @{sbin}/docker-proxy #aa:lint ignore=sbin
 profile docker-proxy @{exec_path} {
   include <abstractions/base>
+  include <abstractions/consoles>
 
   capability net_admin,
   capability net_bind_service,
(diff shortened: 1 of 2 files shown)


apparmor.d/groups/virt/docker-proxy
```diff
@@ -25,6 +27,8 @@ profile docker-proxy @{exec_path} {
   @{PROC}/@{pid}/mountinfo r,
   @{PROC}/@{pid}/cgroup r,
 
+  owner /dev/pts/0 rw,
```
roddhjav: Use the `consoles` abstraction instead.

apparmor.d/profiles-m-r/run-parts
```diff
@@ -15,6 +15,8 @@ profile run-parts @{exec_path} {
   include <abstractions/nameservice-strict>
 
   capability mknod,
+  capability net_admin,
+  capability sys_ptrace,
```
roddhjav: If run-parts is *only* starting `resolved` and `clamav` there is no reason it requires these new capabilities.

skarpinis:
Thanks!

roddhjav:
Thanks, merged.

PR #1056, opened 2026-03-06
 pipewire: remove already included rule

JND94:


roddhjav:
Good catch thanks, merged.

PR #1069, opened 2026-03-17
Draft: initial support for plasma-login-manager

JND94:
TODO:
- dbus rules
- add more abstractions, remove no longer needed rules
- add variable for plasmalogin config, cache and state folders
- add profiles for plasma-login-greeter and plasma-login-wallpaper
- add support for kwin_x11 ?

I'm very clueless about dbus stuff so please take these profiles as a base if they're useful.
Personally, I use them in enforced without issues.

### DIFF
--- a/apparmor.d/groups/kde/ksecretd
+++ b/apparmor.d/groups/kde/ksecretd
@@ -1,0 +1,31 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2025 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/ksecretd
+profile ksecretd @{exec_path} flags=(attach_disconnected,mediate_deleted) {
+  include <abstractions/base>
+  include <abstractions/kde-strict>
+  include <abstractions/graphics>
+
+  @{exec_path} mr,
+
+  owner @{user_config_dirs}/kwalletrc r,
+
+  owner @{user_share_dirs}/kwalletd/ rw,
+  owner @{user_share_dirs}/kwalletd/** rwkl -> @{user_share_dirs}/kwalletd/#@{int},
+
+  owner @{run}/user/@{uid}/kwallet5.socket rw,
+
+  @{PROC}/sys/kernel/random/boot_id r,
+
+  /dev/tty r,
+
+  include if exists <local/ksecretd>
+}
+
+# vim:syntax=apparmor


roddhjav:
Thanks, I too have a draft for these new profiles. However, please note: plasmalogin is a fork of SDDM. So the initial architecture can be very similar to it. It turns out I currently do not like the way the SDDM profiles words (too many things in one big profile). Thus, it is the opportunity to rewrite them, in better. By design these kind of program need to be handled with care too: you have privileged part, auth program, userland part.

To sum up, in your PR:
- `plasmalogin` is good as it does not include (`ix`) plasmalogin-helper (the current sddm profile did).
- `plasmalogin-helper` should also, not include use shell: `@{shells_path} Px -> something,`: it is one of the sensitive part where the auth will take place. This part will later transition to unprivileged process.

JND94:
Thanks for the feedback, I'm not quite sure what you mean here: 

> plasmalogin-helper should also, not include use shell

and how it should be handled 🤔

roddhjav:
I will try to push a draft of what I mean in this PR, on top of your actuall change, so you can test it and edit it too.

curiosityseeker:
FWIW, in my `plasmalogin-helper` profile I'm calling `ksecretd` via Cx : 

```
profile ksecretd flags=(complain) {

     include <abstractions/base>
     include <abstractions/graphics>
     include <abstractions/kde-strict>
     include <abstractions/video>

     @{bin}/ksecretd mr,

     @{bin}/gpgconf Px,

     @{etc_ro}/machine-id r,

     owner @{user_config_dirs}/kwalletrc rw,
     owner @{user_config_dirs}/kwalletrc.lock rwk,

     owner @{user_share_dirs}/kwalletd/ r,
     owner @{user_share_dirs}/kwalletd/#@{int8} rw,
     owner @{user_share_dirs}/kwalletd/kdewallet.kwl rw,
     owner @{user_share_dirs}/kwalletd/kdewallet_attributes.json rw,
     owner @{user_share_dirs}/kwalletd/kdewallet.kwl.@{rand6} rwl -> @{user_share_dirs}/kwalletd/#@{int8},
     owner @{user_share_dirs}/kwalletd/kdewallet_attributes.json.@{rand6} rwl -> @{user_share_dirs}/kwalletd/#@{int8},
     owner @{user_share_dirs}/kwalletd/kdewallet.salt r,

     /dev/tty r,

     owner @{PROC}/@{pid}/fd/ r,
     owner @{PROC}/@{pid}/task/@{tid}/comm w,

     @{sys}/devices/system/cpu/cpu@{int}/cpu_capacity r,

     include if exists <local/plasmalogin-helper_ksecretd>

     }
``

JND94:
@curiosityseeker I think ksecretd should be made with its own separated profile, it doesn't need less permissions with plasmalogin than by its own and there are more display managers

curiosityseeker:
> @curiosityseeker I think ksecretd should be made with its own separated profile, it doesn't need less permissions with plasmalogin than by its own and there are more display managers

You are right. I'm preparing a PR for a ksecretd profile.

curiosityseeker:
Btw. I'm seeing here that `plasmalogin-helper` always tries to execute `pacman` for whatever reason:

`ALLOWED plasmalogin-helper exec @{bin}/pacman -> plasmalogin-helper//null-@{bin}/pacman comm=bash requested_mask=x denied_mask=x`

I've added `deny @{bin}/pacman rx,` to the profile. Strange enough, I'm still getting above request although deny rules should also work in complain mode and should not be logged by default.

curiosityseeker:
> add support for kwin_x11 ?

AFAIK, plasmalogin no longer supports X11 but only wayland.

roddhjav:
In #1078 (based on this PR, so it could be merged here first). It adds:
- ksecretd 
- some plasma-login profile
- Revisit plasmalogin, plasmalogin-helper
- Add plasmalogin-shell

> Btw. I'm seeing here that plasmalogin-helper always tries to execute pacman for whatever reason:

sddm wants the same (as for flatpak). A subprofile is the best solution.

FYI:
- `deny @{bin}/pacman x,` only work on enforced profile (`deny` and `x` is a special case)
- Do not deny this kind of things unless you know exactly why it is needed and how it is used.

JND94:
I merged but I think I didn't did it right, in a cleaner way 😅 I'm not familiar with git

roddhjav:
I megred the PR as it is right now. I think it is still in a good WIP stage. However, in complain mode, it should not break anything. Please report / open PR when you find issues.

@JND94 To ensure a good integration here I think you had to allow the project maintainer pushing to your PR. Meanwhile I think I messed up myself.

Merged in https://github.com/roddhjav/apparmor.d/pull/1078

PR #1071, opened 2026-03-20
Minor updates

skarpinis:
Hi, please check if anything let me know thank you.

### DIFF
--- a/apparmor.d/groups/systemd/systemd-hwdb
+++ b/apparmor.d/groups/systemd/systemd-hwdb
@@ -24,6 +24,8 @@ profile systemd-hwdb @{exec_path} flags=(attach_disconnected,mediate_deleted) {
   /etc/udev/hwdb.bin wl -> @{lib}/udev/#@{int},
   /etc/udev/hwdb.d/{,*} r,
 
+  / r,
+
   owner @{PROC}/@{pid}/stat r,
 
   include if exists <local/systemd-hwdb>


apparmor.d/groups/filesystem/mkswap
```diff
@@ -11,6 +11,7 @@ include <tunables/global>
 profile mkswap @{exec_path} {
   include <abstractions/base>
   include <abstractions/disks-write>
+  include <abstractions/consoles>
```
roddhjav: Nitpicking a bit: can you sort them alphabetically (while still keeping base first)

apparmor.d/groups/freedesktop/xorg
```diff
@@ -123,6 +124,8 @@ profile xorg @{exec_path} flags=(attach_disconnected) {
   @{sys}/devices/**/hid r,
   @{sys}/devices/**/power_supply/**/{type,online} r,
   @{sys}/devices/platform/ r,
+  @{sys}/devices/@{pci}/tile0/gt0/freq0/min_freq r,
```
roddhjav: You can't have `0` here, it must be `@{int}`.

apparmor.d/groups/network/NetworkManager
```diff
@@ -134,6 +134,7 @@ profile NetworkManager @{exec_path} flags=(attach_disconnected) {
   @{run}/udev/data/+platform:* r,         # Identifies onboard devices (laptop/board model, power controllers, thermal sensors)
   @{run}/udev/data/+rfkill:* r,           # Kill switch for wireless devices (Wi-Fi, Bluetooth, NFC) to save power
   @{run}/udev/data/n@{int} r,             # For network interfaces
+  @{run}/udev/data/+ieee80211:phy0 r,     # For wireless interfaces
```
roddhjav: For consistency with existing rule elsewhere, use:

```
  @{run}/udev/data/+ieee80211:* r,        # For Wi-Fi devices, such as wireless network cards and access points.
```

apparmor.d/groups/polkit/polkit-agent-helper
```diff
@@ -27,6 +27,7 @@ profile polkit-agent-helper @{exec_path} flags=(attach_disconnected) {
   network netlink raw,
 
   ptrace read peer=pkttyagent,
+  ptrace read peer=unconfined,
```
roddhjav: ptrace to unconfined peer is not a good thing to allow. In this context, the actual unconfined program might be `systemd`. Thus, use the following instead:

```
  ptrace read peer=@{p_systemd},
```

It resolves to unconfined (unless in FSP mode)

apparmor.d/profiles-g-l/git
```diff
@@ -25,6 +25,7 @@ profile git @{exec_path} flags=(attach_disconnected) {
   network netlink raw,
 
   signal send peer=aurpublish,
+  signal send set=term peer=ssh//&unconfined,
```
roddhjav: That is simply not possible using the current setup.

apparmor.d/groups/filesystem/mkswap
skarpinis: Why I like this project.

roddhjav:
Thanks merged!

PR #1078, opened 2026-03-22
Plasmalogin

roddhjav:
See #1069


curiosityseeker:
The complain flag for plasma-login-wallpaper is missing.

roddhjav:
@curiosityseeker good point, I missed this. I will fix it later

Issue #1072, opened 2026-03-21
Git specific allowance variable


murmelgrumpf:
I think it would be good for git to have an easy way to allow certain directories. I have read from the profile and the documentation (who could be more clear about this part btw) that it mainly uses the @{user_projects_dir}, but this is also used by a bunch of other profiles.

Something like @{git_projects_dir} would be perfect. Otherwise the user has to make another profile just for git to add these

Btw, here are some examples on what I would greenlight personnally. It is mainly config files used by me or programs that also use git to fetch stuff. Two of them I would honestly put in the default git profile since they are widespread. But the main issue is that I would like to have the opportunity to just add a git repo I have somewhere to be allowed without compromising on security elsewhere.

oh-my-zsh (which could also just be in the main git file, since it is a widely used extension for zsh)
```type=AVC msg=audit(1774097503.904:172922): apparmor="DENIED" operation="open" class="file" profile="git" name="/home/murmelgrumpf/.oh-my-zsh/.git/HEAD" pid=286169 comm="git" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000FSUID="murmelgrumpf" OUID="murmelgrumpf"
type=SYSCALL msg=audit(1774097503.904:172922): arch=c000003e syscall=257 success=no exit=-13 a0=ffffffffffffff9c a1=561426d7b580 a2=0 a3=0 items=0 ppid=286168 pid=286169 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts16 ses=2 comm="git" exe="/usr/bin/git" subj=git key=(null)ARCH=x86_64 SYSCALL=openat AUID="murmelgrumpf" UID="murmelgrumpf" GID="murmelgrumpf" EUID="murmelgrumpf" SUID="murmelgrumpf" FSUID="murmelgrumpf" EGID="murmelgrumpf" SGID="murmelgrumpf" FSGID="murmelgrumpf"
type=AVC msg=audit(1774097503.909:172923): apparmor="DENIED" operation="open" class="file" profile="git" name="/home/murmelgrumpf/.oh-my-zsh/.git/HEAD" pid=286170 comm="git" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000FSUID="murmelgrumpf" OUID="murmelgrumpf"

[...]

PR #1079, opened 2026-03-22
feat(profile): git: add user_git_dirs

murmelgrumpf:
See #1072

### DIFF
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -71,6 +71,7 @@
 @{user_ssh_dirs}=@{HOME}/@{XDG_SSH_DIR} @{MOUNTS}/@{XDG_SSH_DIR}
 @{user_passwordstore_dirs}=@{HOME}/@{XDG_PASSWORDSTORE_DIR} @{MOUNTS}/@{XDG_PASSWORDSTORE_DIR}
 @{user_private_dirs}=@{HOME}/@{XDG_PRIVATE_DIR} @{MOUNTS}/@{XDG_PRIVATE_DIR}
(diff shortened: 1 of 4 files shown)


murmelgrumpf:
Hey,
@roddhjav I tried to implement it close to the already existing variables. Let me know if there is anything you think should be changed

murmelgrumpf:
Oh and in general. How do I build/test it. The changes were not that big so I could test it manually, but is it explained somewhere? Couldn't find it in the readme or in the docs

apparmor.d/tunables/home.d/apparmor.d
```diff
@@ -71,6 +71,7 @@
 @{user_ssh_dirs}=@{HOME}/@{XDG_SSH_DIR} @{MOUNTS}/@{XDG_SSH_DIR}
 @{user_passwordstore_dirs}=@{HOME}/@{XDG_PASSWORDSTORE_DIR} @{MOUNTS}/@{XDG_PASSWORDSTORE_DIR}
 @{user_private_dirs}=@{HOME}/@{XDG_PRIVATE_DIR} @{MOUNTS}/@{XDG_PRIVATE_DIR}
+@{user_git_dirs}=""
```
roddhjav, line 74: This cannot be set this way. If it is set to an empty value, thus `@{user_git_dirs}/** rwkl` would be an equivalent to `/** rwkl,` and you do not want that.

Like the other variables defined in this file, add 
```
@{XDG_GIT_DIR}=...
```
Again, it cannot be empty, it must be initialized

Then:
```
@{user_git_dirs}=@{HOME}/@{XDG_GIT_DIR} @{MOUNTS}/@{XDG_GIT_DIR}
```

PR #1100, opened 2026-04-16
remove unconfined fallbacks

valoq:
Many cases of unconfined execute can be removed.
Especially problematic is the use of child-open-any in browsers, which effectively allows unconfined shells. If I see this correctly that alone means automatic sandbox escape

### DIFF
--- a/apparmor.d/abstractions/app/chromium
+++ b/apparmor.d/abstractions/app/chromium
@@ -85,7 +85,7 @@
   @{bin}/xdg-email          Px,
   @{bin}/xdg-icon-resource  ix,
   @{bin}/xdg-mime          rix,
-  @{bin}/xdg-open           Px -> child-open-any,
+  @{bin}/xdg-open          rPx -> child-open,
   @{bin}/xdg-settings      rix,
 
   # Installing/removing extensions, applications, and stacked xdg menus
--- a/apparmor.d/groups/browsers/firefox
+++ b/apparmor.d/groups/browsers/firefox
@@ -63,7 +63,7 @@ profile firefox @{exec_path} flags=(attach_disconnected) {
   @{bin}/update-mime-database                        rPx,
   @{lib}/gvfsd-metadata                              rPx,
   @{lib}/mozilla/kmozillahelper                     rPUx,
-  @{open_path}                                       rPx -> child-open-any,
+  @{open_path}                                       rPx -> child-open,
 
   # Common extensions
   @{bin}/browserpass                                           rPx,


roddhjav:
This won't work. It you only allow `Px` instead of `PUx` it means the target profile must exist. In most of these case they do not exsit, thus it would break the profile.

Same for `child-open` instead of `child-open-any`. These profiles need to be able to open anything.

valoq:
https://github.com/roddhjav/apparmor.d/blob/c2056759fd61b5752d090425547e411e9a63d234/apparmor.d/groups/children/child-open-any#L19

in case of child-open-any, this means trivial sandbox escape since anything in bin can be called unconfined, including shells.

valoq:
I updated the patch to only address the child-open-any

roddhjav:
Sadly it won't work either. `child-open` used to be used here, but it has been changed to `child-open-any` because it was to strict in a way it is hard (i.e. as of today, not possible) to relax it: a browser can use any program to open resources.

> in case of child-open-any, this means trivial sandbox escape since anything in bin can be called unconfined, including shells.

The child-open solution is limited and cannot be used everywhere, so we do not have alternative here. Unless, we ensure child-open include the full list of possible GUI app (spoiler alert, unless all apps are in flatpak/snap, we can't).

Besanon:
I asked myself the same question.
I have replaced PUx mit Px in all child* profiles, in abstractions/app-launcher-* , in abstractions/app-open, x11-xsession and replaced in xdg-open the PUx statements with
  @{bin}/dbus-send            Cx -> bus,
  @{bin}/gdbus                Cx -> bus,
  @{bin}/xprop                Px,
  @{bin}/xdg-mime             Px,
  @{bin}/{,e}grep             rix,
I have choosen kwin as windows manager for its apparently better than openbox mit all these PUx lines.
I use Arch Linux with wayland/x running LXQt with kwin_wayland and Apparmor 4.1.7 + apparmor.d.enforced. Whole desktop with some programs - no flatpak, docker or snap.
I needed some adaption in cron and with brave.The start script of brave wasnt confined, so i wrote a simple profile for it.
It took me some hours but i see no difference in the logs up to now, no sluggish behaviour, same speed. I think it was helpful to set these PUx definitions in order to get the whole picture of what is needed in the whole project. 
But actually there are many many PUx lines which can be altered at least in some environments. To my surprise even the dbus* profiles can be altered to avoid PUx. Maybe i encounter still unconfined programs and will see some errors later. 
Maybe its all too bold to set these changes, i certainly missed some point but i wanted to know whats happening. Just to let you now

valoq:
With child-open-any the primary path for exploits is completely open. In most poc exploits, the attacker would not even notice that firefox is confined because the most common way to use code execution is just to run the payload in a shell when available. I don't see what purpose these profiles could provide with this permissions, they seem utterly useless. If instead child-open were to be used, there may be some rare cases where the confinement gets in the way, but it still has a purpose. Instead of using child-open-any the more honest alternative would be to disable these profiles. Or accept the unavoidable restriction and have a useful profile.

roddhjav:
`child-open-any` could be restricted a bit; however, let me be clear: I am not going to provide a profile that knowingly breaks normal usage of a program. That is against everything this project stands for. Not to mention the fact that I am the one who has to reply (for free) to legitimately angry users who are wondering why their browser broke.

Note: this could potentially be restricted further depending on the security level implemented: https://apparmor.pujol.io/security/architecture/#security-levels

> I don't see what purpose these profiles could provide with this permissions, they seem utterly useless. 

Here, you are assuming the attacker purposely know the apparmor profile and wants to bypass it. It is a concern obviously, however as of today most attacks/security vulnerabilities in wild are oblivious of MAC restriction.

valoq:
Actually I was trying to argue that the permissions allow bypass using a primary execution path that would mean the attacker does not know the process is confined.
Starting a shell is common after code execution.

I feel like there should be at least an indicator when a profile has known escape issues, maybe in the header of the profile so users know what to expect. The security level is a future solution too.

roddhjav:
Things like shell or coreutils escape (kind of obvious non GUI app) are within what I meant by "child-open-any could be restricted a bit".

> Starting a shell is common after code execution.

True, but when done with `child-open*` you need to use an open helper first. With a direct shell execution from the profile, you would stay in the profile.

curiosityseeker:
> however, let me be clear: I am not going to provide a profile that knowingly breaks normal usage of a program. That is against everything this project stands for. Not to mention the fact that I am the one who has to reply (for free) to legitimately angry users who are wondering why their browser broke.

I can understand this. However:
1. Every security technology has to balance usability and security. In this case I agree with @valoq that using `child-open-any` is basically a built-in sandbox escape, and this should definitely be avoided.
2. The question is why should, e.g., a critical appalication like Firefox be able to execute every random application on my system? The most important and mostly used applications are covered in `child-open`, and many of them are profiled. If important ones are missing, then, by all means, let's add them to the list and write profiles for them if needed.
3.  If users really need Firefox to be able to execute an "exotic" application we can expect them to create a rule in `local` on their own. After all, we also expect users to [personalize](https://apparmor.pujol.io/configuration/) this project to their own needs (and I guess this is more difficult for most users than writing their own `ux,` rule).

> Here, you are assuming the attacker purposely know the apparmor profile and wants to bypass it. It is a concern obviously, however as of today most attacks/security vulnerabilities in wild are oblivious of MAC restriction.

But this is basically "security by obscurity" and we should not rely in that.

roddhjav:
So, I am currently testing a few things. But first, a bit of theory:

- Firefox needs to be able to open any UI app because for instance you can open a Spotify link directly in Spotify, a matrix link using any possible matrix client...

- `child-open` aims to be quite restrictive: only labelled app (and sandbox manager) should be allowed. However, it is hard to label everything. This is why at some point I gave up and moved to `child-open-any` for the most demanding profile.

- `child-open-any`  aims to open all **GUI** app. As of today it allows everything. We could easily restrict it and `deny` some well-know cli app, but that would not be really the MAC way of doing things (not to mention the side effects). However, there is a possible solution: the list of GUI app with a profile can easily be generated. It would still be limited to app with profile, but it is a starting point.

curiosityseeker:
@roddhjav : Thanks, this sounds promising, indeed :+1:

Besanon:
yes, indeed promising, please invest some effort in this. You have a close look at the profiles to narrow the chances of exploits. Maybe a look at Mikhail Kurinnois apparmor profiles shows how root and user shells could be set up.

curiosityseeker:
>  Maybe a look at Mikhail Kurinnois apparmor profiles shows how root and user shells could be set up.

I couldn't find anything related. Typo?

Besanon:
sorry, its a bit hidden, please have a look at:
https://github.com/viewizard/gentoo-apparmor/tree/master/apparmor.d
He has recently done some work on it.
Its working with sysv runit (some mods) and openrc. I tried to  replace some ux in sh_shells with Px -> shell_root of the profiles 1_shell. Needs some adaptation but i think one could consider it to replace the ux in the profiles one after another
Or am i wrong? Overlooked something bad...?

PR #1102, opened 2026-04-24
fix(abstraction): Avoid conflicting exec rules.

emixam16:
Test abstraction allows executing /tmp/*test*/** to support test suites. However, some profiles, like do-release-upgrade also perform explicit Cx transitions.

If the executed file matches the regex, it will fail with profile xxx has merged rule xxx with conflicting x modifiers.

Fixed by adding priority=-1 in the abstraction to allow smooth overrides.


roddhjav:
Good catch thanks.

PR #1105, opened 2026-04-30
Update cups-notifier-dbus and ippfind flags

curiosityseeker:
profile cups-notifier-dbus flags=(attach_disconnected) { run/dbus/system_bus_socket r, # Failed name lookup - disconnected path

profile cups-notifier-dbus flags=(attach_disconnected) { run/dbus/system_bus_socket r, # Failed name lookup - disconnected path

/run/dbus/system_bus_socket is already covered by the `bus-system` abstraction.

### DIFF
--- a/apparmor.d/groups/cups/cups-notifier-dbus
+++ b/apparmor.d/groups/cups/cups-notifier-dbus
@@ -7,7 +7,7 @@ abi <abi/4.0>,
 include <tunables/global>
 
 @{exec_path} = @{lib}/cups/notifier/dbus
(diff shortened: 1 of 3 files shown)


roddhjav:
This need to be changed in the profile directly. Now, `dists/flags/main.flags` is only used to overwrite the profile the mode (complain, enforce...), not all the flags anymore.

curiosityseeker:
I see! Changed accordingly.

roddhjav:
Thanks, merged!

PR #1112, opened 2026-05-13
fix(profile): missing ansible abstraction for ssh

### LINKED ISSUE #1111: Ansible handling
```
...
apparmor="DENIED" operation="mknod" class="file" profile="ssh" name="/home/zar3bski/.ansible/cp/b00755bc79.RrMzPOJWr6x5Di3I"  comm="ssh" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000 FSUID="zar3bski" OUID="zar3bski"
apparmor="DENIED" operation="mknod" class="file" profile="ssh" name="/home/zar3bski/.ansible/cp/b00755bc79.qB7uwo7aJpSbU6tf"  comm="ssh" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000 FSUID="zar3bski" OUID="zar3bski"
```

After checking the `ssh` profile, it seem that tuning `XDG_SSH_DIR` won't make the trick. The following local inclusion solves the problem

```
owner @{HOME}/.ansible/cp/** rwl,
```

Technically speaking, I fixed my problem but I create the issuer i) for future references ii) to know if/how this could be included into **apparmor.d** and participate to the discussion. Since [ansible](https://docs.ansible.com/) is a pretty common, the need is likely to emerge quite often. 

About the possible implementation of the feature, I wondered. Since `ansible` quite often lives into python virtual environments, it is tricky to predict `ansible` path and handle the case with child process directives. The alternative would be to edit, like I did, the `ssh` profile but this would assume the existence of an tool independent from `ssh`  and I'm not sure if this would be aligned with  **apparmor.d** philosophy. Anyway, let me know with way you would like to go and I'll submit a PR

zar3bski:
A slightly better solution than the one I suggested in #1111 : the abstraction existed but needed: 
- access to `@{HOME}/.ansible/cp`
- to be included in the `ssh` profile

close #1111

### DIFF
--- a/apparmor.d/abstractions/ansible
+++ b/apparmor.d/abstractions/ansible
@@ -5,6 +5,7 @@
   abi <abi/4.0>,
 
   owner @{HOME}/.ansible/tmp/ansible-tmp-*/* rw,
+  owner @{HOME}/.ansible/cp/** rwl,
 
   include if exists <abstractions/ansible.d>
 
--- a/apparmor.d/groups/ssh/ssh
+++ b/apparmor.d/groups/ssh/ssh
@@ -9,6 +9,7 @@ include <tunables/global>
 
 @{exec_path} = @{bin}/ssh
 profile ssh @{exec_path} flags=(attach_disconnected) {
+  include <abstractions/ansible>
   include <abstractions/base>
   include <abstractions/consoles>
   include <abstractions/devices-u2f>


roddhjav:
I am not adding this as default to the ssh profile. In a few month it will be conditionally added in this profile as long as ansible is installed. However, for now you are expected to add it to your local addition.

zar3bski:
Ok, closing the related issue

PR #1126, opened 2026-06-03
fix(profile): gpu-screen-recorder revert 471f35c

JND94:
I'm sorry, it was a bug in arch's packaging
I assumed wrong without a proper research, it will not happen again


JND94:
network required for streaming, still missing webcam, probably just need camera abstraction

Issue #24, opened 2022-01-22
common policy


beroal:
Hi. I'm new to AppArmor. Writing a profile surely requires profound knowledge of Linux internals. Reading your profiles was of great help to me. Is there a written approach of dealing with permissions (what is acceptable and what is not) adopted by this repository? Examples of what I mean:

- `/etc/machine-id`. On one hand, it can be used to track a computer. On the other hand, I believe, there are plenty of ways to track using writing access to common cache directories.
- Consoles. AFAIK, a program having access to consoles can read from all consoles, not just its own `stdin`. Hence it will be able to read a password, for example.
- `dconf`. I see many GUI programs require it. This is a configuration server like Windows Registry. Does the access to `dconf` gives a way to modify other programs' settings? That would be unacceptable. Configuration files don't have this vulnerability.

Knowing your policy will be useful if I ever contribute to your repository and for writing my own profiles.

roddhjav:
Hi.

There is no written approach of dealing with permissions. I can only see of a few rules: 
1. As these are mandatory access control policies only what it explicitly required should be authorized. Meaning, you should not allow everything (or a large area) and blacklist some sub area.
2. A **should not break a normal usage of the confined software**. It can be complex as simply running the program for your own use case is not always exhaustive of the program features and required permissions.
3. Try to respect the profile guideline: [CONTRIBUTING.md](https://github.com/roddhjav/apparmor.d/blob/master/CONTRIBUTING.md), and have a look at the links at the end of this file too.

Regarding `machine-id` and `dconf`, the first thing is that most of the time denying them would break the normal use of the software, so you just can't.

More generally, for `machine-id`, this is an example of privacy feature that is simply not provided by a MAC policy as it can restrict the access of a file, not its content. If you care about it, it is common to use the machine-id shared across all the anonymity distribution such as Tail and Whonix. See https://github.com/Whonix/dist-base-files/blob/master/etc/machine-id

beroal:
> Regarding `machine-id` and `dconf`, the first thing is that most of the time denying them would break the normal use of the software, so you just can't.

Should we use the `dconf` abstraction?

roddhjav:
Yes, when available you should always use abstraction anyway. Now, a few profile require more than what is in the `dconf` abstraction so both abstraction and in-line rules are used.

nobodysu:
- It's also possible to regenerate the id on boot with two sequential commands:
```
# dbus-uuidgen --ensure=/etc/machine-id
[...]

PR #1129, opened 2026-06-07
feat(profile): gsr-ui shell subprofile, file_inherit denials

JND94:
I'm still unsure about /etc/machine-id, it doesn't seem to break normal usage yet what I can gather from #24 is that we probably should just allow it and privacy aware users should handle it by their own


roddhjav:
Thanks, merged.

Note: machine-id is most likely required by dbus. In this case, it is the dbus-session abstraction that is needed.

PR #1143, opened 2026-06-24
fix(profile): update systemd-logind

JND94:
new denials after 261 update
`DENIED  systemd-logind open @{sys}/firmware@{efi}/efivars/LoaderTpm2ActivePcrBanks-@{uuid} comm=systemd-logind requested_mask=r denied_mask=r
DENIED  systemd-logind open @{sys}/firmware@{efi}/efivars/StubPcrKernelImage-@{uuid} comm=systemd-logind requested_mask=r denied_mask=r`

Should /efi/ be @{efi} in the rules ?


roddhjav:
Thanks, merged.

> Should /efi/ be @{efi} in the rules ?

No, /efi/ does not depend on the system here. It is not the location of the EFI partition.

PR #1148, opened 2026-07-06
fix(profile): xdg-mime allow reading the MIME database via abstractions/mime

ParadaCarleton:
Include `abstractions/mime`, which grants read access to both the system (`@{system_share_dirs}/mime/{,**}`) and per-user (`@{user_share_dirs}/mime/{,**}`) MIME databases plus the `mimeapps.list` variants. This is the same abstraction already used by other MIME-consuming profiles (`evince-thumbnailer`, `fwupd`, `syncthing`, `appstreamcli`).

Verified on Arch/CachyOS with `apparmor.d.enforced`: after adding the include, `xdg-mime query default text/plain` returns the correct handler with no denials.


roddhjav:
Thanks, merged

PR #1151, opened 2026-07-06
fix(profile): gnome-extension-gsconnect attach the browser nativeMessagingHost entrypoint

ParadaCarleton:
The `gnome-extension-gsconnect` profile's `@{exec_path}` only lists `service/daemon.js` and `gsconnect-preferences`. It does not include `service/nativeMessagingHost.js`, the entrypoint a browser launches to speak the GSConnect native-messaging protocol.

Because that path isn't in the attach path, when a browser (via a `Px` transition) launches the native-messaging host, it doesn't attach to this profile — under enforce mode it falls to a null/denied child profile, breaking the browser <-> GSConnect integration.

This adds `@{share_dirs}/service/nativeMessagingHost.js` to `@{exec_path}` (same variable/style as the existing entries) so the browser `Px` transition lands on this profile correctly.

### DIFF
--- a/apparmor.d/abstractions/app/chromium
+++ b/apparmor.d/abstractions/app/chromium
@@ -100,6 +100,8 @@
   # Gnome shell integration
   @{bin}/chrome-gnome-shell            Px,
   @{bin}/gnome-browser-connector-host  Px,
+  @{user_share_dirs}/gnome-shell/extensions/gsconnect@andyholmes.github.io/service/nativeMessagingHost.js  Px -> gnome-extension-gsconnect-native,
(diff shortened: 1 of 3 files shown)


roddhjav:
Thanks, however, `nativeMessagingHost.js` is not the `gsconnect` extension. It is a small dbus script that should be confined in a separate profile. So it is not the solution here.

>  when a browser (via a Px transition) launches the native-messaging host,

This is the real issue: there is no rule that allow transition to the native tool.

Please share the apparmor logs you had.

ParadaCarleton:
My bad, closing this one.

PR #1152, opened 2026-07-06
fix(profile): gnome-browser-connector-host allow gi typelibs, gio modules and user gschemas

ParadaCarleton:
Under enforce mode, `gnome-browser-connector-host` (a PyGObject / Python script) is
denied several reads it needs at runtime:

- **girepository typelibs** (`@{lib}/girepository-1.0/`) — required by any `import gi`
  / PyGObject code.
- **gio modules** (`@{lib}/gio/modules/`, `giomodule.cache`) — loaded by GLib/GIO at
  startup.
- **the per-user compiled gschemas** (`@{user_share_dirs}/glib-2.0/schemas/gschemas.compiled`)
  — `abstractions/gschemas` only grants the **system** `gschemas.compiled`, not the
  per-user one, so the connector host is denied when a user schema override exists.
- **`@{HOME}/`** directory read.

This adds the missing reads.

Note: the per-user `gschemas.compiled` line is arguably better placed in
`abstractions/gschemas` itself, which would fix the whole class of GLib/GIO apps
project-wide rather than just this profile. Happy to move it there instead if you
prefer — flagging it as an alternative.

### DIFF
--- a/apparmor.d/groups/gnome/gnome-browser-connector-host
+++ b/apparmor.d/groups/gnome/gnome-browser-connector-host
@@ -20,6 +20,12 @@ profile gnome-browser-connector-host @{exec_path} {
 
   @{lib}/@{python_name}/site-packages/gnome_browser_connector/__pycache__/{,**} rw,
 
+  @{lib}/girepository-1.0/{,*}                                 r,
+  @{lib}/gio/modules/{,*.so*}                                  rm,
+  @{lib}/gio/modules/giomodule.cache                           r,
+  owner @{user_share_dirs}/glib-2.0/schemas/gschemas.compiled  r,
+  owner @{HOME}/                                               r,
+
   owner @{PROC}/@{pid}/mounts r,
 
   include if exists <local/gnome-browser-connector-host>


roddhjav:
Could you provide the apparmor logs you had with this issue.

Note: gio and girepository libs should already be included by the base-strict abstraction with:
```
  @{lib}/** r,
  @{lib}/**.so* m,
```

ParadaCarleton:
You're right on both counts. The profile already pulls in `abstractions/base-strict`, which provides:

```
@{lib}/** r,
@{lib}/**.so* m,
```

so the `girepository-1.0` and `gio/modules` lines here are redundant. And `abstractions/gschemas` is already included too, so the single genuine gap is the **per-user** compiled schemas (`~/.local/share/glib-2.0/schemas/gschemas.compiled`) — which is exactly what #1158 adds to the `gschemas` abstraction. Since `gnome-browser-connector-host` includes that abstraction, #1158 covers this case with nothing left over.

On the logs: every event was attributed to the stacked complain label `firefox//null-/usr/bin/gnome-browser-connector-host` (~5.4k of them) — **0 ever reached the real `gnome-browser-connector-host` profile**, because the browser was in complain mode so the `Px` never enforced. The gaps I saw under that label were `getattr`/`open` on `/usr/lib/girepository-1.0/` and the user `gschemas.compiled`; once the transition enforces, `base-strict` + #1158 handle them.

I'll close this in favour of #1158.

PR #1154, opened 2026-07-06
fix(profile): gsd-xsettings allow querying the fcitx5 input method

ParadaCarleton:
`gsd-xsettings` queries the active input method at session startup. It already handles IBus via `@{lib}/{,ibus/}ibus-x11 rPx`, but has no rule for fcitx5, a mainstream IBus alternative. Under enforce mode, `gsd-xsettings` invoking `fcitx5-remote` is denied.

This adds a rule mirroring the existing ibus-x11 handling:

```
  @{bin}/fcitx5-remote  rPUx,
```

Since there is no dedicated fcitx5 profile in the tree, `rPUx` is used so the binary transitions to a profile if one ever exists and otherwise runs unconfined, matching the intent of the ibus-x11 line without introducing a dangling profile reference.

Note this is niche — it only affects fcitx5 users. Verified on apparmor.d.enforced.


roddhjav:
Thanks, merged.

PR #1156, opened 2026-07-06
fix(nautilus): allow reading /etc/os-release and /proc/diskstats

ParadaCarleton:
Under enforce mode the `nautilus` profile denies two read-only accesses it makes in normal use:

```
apparmor="DENIED" operation="open" profile="nautilus" name="/etc/os-release" requested_mask="r"
apparmor="DENIED" operation="open" profile="nautilus" name="/proc/diskstats" requested_mask="r"
```

- `/etc/os-release` — distro name/logo shown in the file-manager About and volume Properties UI.
- `/proc/diskstats` — disk activity for the mounted-volume / Trash views.

This adds both as read-only rules next to the existing `/etc/fstab` and `@{PROC}` entries. `@{PROC}/diskstats r,` matches the idiom already used by the `htop`, `vmstat`, `sysstat-sadc` and `cockpit` profiles.

Tested on `apparmor.d.enforced` (Arch): both denials disappear, no regressions.

### DIFF
--- a/apparmor.d/groups/gnome/nautilus
+++ b/apparmor.d/groups/gnome/nautilus
@@ -91,7 +91,8 @@ profile nautilus @{exec_path} flags=(attach_disconnected) {
   /usr/share/thumbnailers/{,**} r,
   /usr/share/tracker*/{,**} r,
 
-  /etc/fstab r,
+  /etc/fstab      r,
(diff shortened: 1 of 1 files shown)


roddhjav:
Thanks, merged

PR #1159, opened 2026-07-06
fix(borg): allow reading @{PROC}/@{pid}/mounts

ParadaCarleton:
BorgBackup reads its own `/proc/<pid>/mounts` to enumerate mount points (e.g. for `--one-file-system` boundaries and to skip pseudo-filesystems). Under enforce:

```
apparmor="DENIED" operation="open" profile="borg" name="/proc/1234/mounts" requested_mask="r"
```

Adds it owner-scoped next to the existing `@{PROC}/@{pid}` reads. Tested on apparmor.d.enforced.


roddhjav:
Thanks, merged.

PR #1160, opened 2026-07-06
fix(gtk-strict, ibus-strict): allow reading X Compose files

ParadaCarleton:
GTK and ibus read the user Compose file `~/.XCompose` and maintain a compiled Compose cache under their cache dirs (`~/.cache/gtk-3.0/compose/`, `~/.cache/ibus/compose/`). Both are denied under the strict abstractions:

```
apparmor="DENIED" profile="ibus-engine-simple" name="/home/user/.XCompose" ...
apparmor="DENIED" profile="firefox" name="/home/user/.cache/gtk-3.0/compose/..." ...
```

Allow reading `~/.XCompose` and read/write of the compose caches.


roddhjav:
Again, share the full raw logs, this touches core abstraction. There are not obvious changes.

PR #1161, opened 2026-07-06
fix(geoclue): allow reading /etc/os-release

ParadaCarleton:
geoclue reads `/etc/os-release` to compose its User-Agent for the online location provider. The read is denied:

```
apparmor="DENIED" operation="open" profile="geoclue" name="/etc/os-release" ...
```

https://claude.ai/code/session_01DohXyQUuSV3iMBjybBg5m8


roddhjav:
Thanks, merged

PR #1167, opened 2026-07-08
feat(profile): gsr add controller hotkeys, systemctl subprofile, more file_inherit denials

JND94:


apparmor.d/profiles-g-l/gsr-ui
```diff
@@ -28,10 +30,13 @@ profile gsr-ui @{exec_path} {
   @{bin}/gsr-kwin-helper     rPx,
   @{bin}/gsr-notify          rix,
   @{bin}/gpu-screen-recorder rPx,
-  @{bin}/systemctl           rPx -> child-systemctl,
+  @{bin}/systemctl            cx -> systemctl,
```
roddhjav:
```suggestion
  @{bin}/systemctl            Cx -> systemctl,
```

apparmor.d/profiles-g-l/gpu-screen-recorder
```diff
@@ -45,6 +45,8 @@ profile gpu-screen-recorder @{exec_path} {
 
   /usr/share/gsr-ui/fonts/{,**} r,
 
+  deny /dev/input/event@{int} r, # file_inherit
```
roddhjav: Are you **sure** the program does not need access to input devices?

- If it does: use the `input` abstraction instead.
- If it is really a file that was not closed by the parent program, move the deny rules at the end of the profile.

It goes the same for similar rules in this PR.

apparmor.d/profiles-g-l/gsr-ui
```diff
@@ -50,13 +55,26 @@ profile gsr-ui @{exec_path} {
...
     deny /dev/dri/card@{int} r, # file_inherit
+    deny /dev/input/event@{int} r, # file_inherit
 
     include if exists <local/gsr-ui_shell>
   }
 
+  profile systemctl flags=(complain) {
+    include <abstractions/base>
+
+    @{bin}/systemctl r,
```
roddhjav:
```suggestion
    include <abstractions/base>
    include <abstractions/app/systemctl>

```

apparmor.d/profiles-g-l/gpu-screen-recorder
JND94: I'm fairly sure, only gsr-global-hotkeys or gsr-ui (in this case just for gamepads) should be able to read input, then control gpu-screen-recorder
`input` abstraction is probably too broad for this case, program has predefined keys that cannot be changed, we don't need joystick nor mouse access

roddhjav:
Thanks, merged!

JND94:
@roddhjav ping

PR #1169, opened 2026-07-09
Fix calibre

CossackLucas:
Official release for Linux ([here](https://calibre-ebook.com/download_linux)) is installed in /opt as default. Additionally, at least from version 9.6 onwards temp files names include word characters instead of just alphanumeric


apparmor.d/profiles-a-f/calibre
```diff
@@ -87,7 +90,7 @@ profile calibre @{exec_path} {
   owner @{user_cache_dirs}/calibre/ rw,
   owner @{user_cache_dirs}/calibre/** rwkl -> @{user_cache_dirs}/calibre/**,
 
-  owner @{tmp}/@{rand8} rw,
+  owner @{tmp}/@{w}@{w}@{w}@{w}@{w}@{w}@{w}@{w} rwl,
```
roddhjav:
```suggestion
  owner @{tmp}/@{word8} wl,
```

CossackLucas: Agreed, I must've just copied it from a fast fix in my local overrides

CossackLucas (commented):
Agreed on changes

roddhjav:
Thanks, merged!

PR #1171, opened 2026-07-10
fix(plasma-login-manager) Adding of missing permissions on Plasma 6.7

CossackLucas:
Fixes locking PLM on Plasma 6.7 from running.
Includes changes related to proprietary Nvidia drivers. This cache is used by other Plasma components, but so far I was adding the same entries separately each time.

The only issue I have is with vapoursynth - it does not seem to me to be connected to plasmalogin-helper and I haven't seen any problem if denied, but i left it in the profile as is requested.

Logs are in a file, as the first check generated over 2k lines for all 8 profiles.
[Logs.txt](https://github.com/user-attachments/files/29886347/fix-plm.txt)


roddhjav:
Thanks a lot. Merged!

PR #1173, opened 2026-07-11
Draft: AppImage profiles

JND94:
AppRun as how it is right now it's a base profile that needs further and broad rules to accomodate all kind of apps that are shipped as AppImage, for now I mark this as draft, users can adapt AppRun to suit their own needs.

Profiles assume appimages are stored somewhere in @{HOME}

### DIFF
--- a/apparmor.d/profiles-a-f/AppImage
+++ b/apparmor.d/profiles-a-f/AppImage
@@ -1,0 +1,54 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2022-2024 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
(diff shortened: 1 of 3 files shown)


JND94:
I surrender, works fine locally (without builder)

roddhjav:
The build error:

> ✗ Error: userspace .build/enforce/apparmor.d/AppRun: variable rand6 not defined

It is because you use `@{rand6}` in the definition of `@{exec_path}`. The prebuild system parses this variable to bypass some apparmor upstream restriction. However, this prebuilds only have a few variables defined by default. For rand6 it will have to be added here: https://github.com/roddhjav/apparmor.d/blob/e81127162be8ca2f1bbfbcce917bccdc1897709a/pkg/aa/apparmor.go#L85

roddhjav:
I agree that AppImage should be fully supported by apparmor.d. However, beware, it is a long and complex job as you are not confining one program but any program that AppImage can handle.

In spirit, it should thus be similar to what I have done for flatpak app, See https://github.com/roddhjav/apparmor.d/tree/main/apparmor.d/groups/flatpak

PR #1177, opened 2026-07-14
feat(checkupdates) Introduction of a profile for the checkupdates script for pacman (Arch)

CossackLucas:
Profile for checkupdates script. The script is recommended for checking updates on ArchLinux and other distros using pacman.

Small changes required in the pacman profile, as the database is created by non root user.

[Logs](https://github.com/user-attachments/files/30016884/aa-profile-pacman-checkupdates.txt)


roddhjav:
Thanks, this looks good, merged.

PR #1178, opened 2026-07-16
Draft: calibre - profile's generalization for binary install and introducing poppler's profile

CossackLucas:
As ArchLinux finally has current releases of calibre in the official repo, I was able to compare it with the official binary release. Apart from noticing not including abstraction for chromium, a few things required finding nonstandard solutions and I'd like to check them before continuing.

1. @{app_bin} - using it generalized previous solution (my 1st version had problems with additional apps, regular calibre worked fine)
2. @{calibre_bin} - gives flexibility if default is changed and/or we find out that some distro is installing binary version. Additionally, it could be moved somewhere in tunables, to allow adding installation directory by the users, if they install it somewhere else (but it does seem like an overkill)
3. Pdf related apps - instead of using regular profiles, all 3 in my concept use one child profile. Seemed simpler than giving them general access to /tmp for those without need of it or creating 3 separate child profiles with minimal differences.
4. Binary version of calibre includes a version of poppler library so for this situation using of included libraries was allowed

### DIFF
--- a/apparmor.d/profiles-a-f/calibre
+++ b/apparmor.d/profiles-a-f/calibre
@@ -7,14 +7,18 @@ abi <abi/4.0>,
 
 include <tunables/global>
 
-@{exec_path}  = @{bin}/calibre{,-*} @{bin}/calibredb @{bin}/ebook{,-*}
-@{exec_path} += @{bin}/fetch-ebook-metadata
-@{exec_path} += @{bin}/lrs2lrf @{bin}/lrf2lrs @{bin}/lrfviewer @{bin}/web2disk
-@{exec_path} += /opt/calibre{,-*}{,/bin}/calibre{,-*}
+@{domain} = org.chromium.Chromium
+
+@{bin_dirs} = @{bin} /opt/calibre{,/bin}
+
+@{exec_path}  = @{bin_dirs}/calibre{,-*} @{bin_dirs}/calibredb @{bin_dirs}/ebook{,-*}
+@{exec_path} += @{bin_dirs}/fetch-ebook-metadata
+@{exec_path} += @{bin_dirs}/lrs2lrf @{bin_dirs}/lrf2lrs @{bin_dirs}/lrfviewer @{bin_dirs}/web2disk
 profile calibre @{exec_path} {
   include <abstractions/base>
   include <abstractions/bus/session/org.kde.StatusNotifierWatcher>
   include <abstractions/bus/system/org.freedesktop.UDisks2>
+  include <abstractions/common/chromium>
   include <abstractions/dconf-write>
   include <abstractions/desktop>
   include <abstractions/devices-usb>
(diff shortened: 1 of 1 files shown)


apparmor.d/profiles-a-f/calibre
```diff
@@ -7,14 +7,21 @@ abi <abi/4.0>,
...
+@{app_bin} = @{bin} @{calibre_bin}{,/bin}
+
+@{exec_path}  = @{app_bin}/calibre{,-*} @{app_bin}/calibredb @{app_bin}/ebook{,-*}
+@{exec_path} += @{app_bin}/fetch-ebook-metadata
+@{exec_path} += @{app_bin}/lrs2lrf @{app_bin}/lrf2lrs @{app_bin}/lrfviewer @{app_bin}/web2disk
 profile calibre @{exec_path} {
   include <abstractions/base>
   include <abstractions/bus/session/org.kde.StatusNotifierWatcher>
   include <abstractions/bus/system/org.freedesktop.UDisks2>
+  include <abstractions/common/chromium>
```
roddhjav, line 21: Calibre is not based on Chromium nor an Electron app, thus you cannot use this abstraction. Considering the size of this abstraction, please review the full profile without it.

apparmor.d/profiles-a-f/calibre
```diff
@@ -7,14 +7,21 @@ abi <abi/4.0>,
...
-@{exec_path}  = @{bin}/calibre{,-*} @{bin}/calibredb @{bin}/ebook{,-*}
-@{exec_path} += @{bin}/fetch-ebook-metadata
-@{exec_path} += @{bin}/lrs2lrf @{bin}/lrf2lrs @{bin}/lrfviewer @{bin}/web2disk
-@{exec_path} += /opt/calibre{,-*}{,/bin}/calibre{,-*}
+@{domain} = org.chromium.Chromium
+
+# could be a tunable to allow users change, if they don't install it in default dir
+@{calibre_bin} = /opt/calibre
+
+@{app_bin} = @{bin} @{calibre_bin}{,/bin}
```
roddhjav: To have the name naming across profiles, this should be named `@{bin_dirs}`.

`/opt/calibre` should be fully part of it, not in its own var.

apparmor.d/profiles-a-f/calibre
CossackLucas: That one is hard to avoid - calibre uses qtwebengine, chromium based framework, at least as an ebook viewer. And as it's spawned with no new privileges flag I see now way avoiding including it. I tried to run it in a separate profile, but that's how I found out the usage of the flag.

CossackLucas:
Checks finished.
Unfortunately QtWebEngine is used everywhere, where html-based ebook formats are viewed (including preview in the book editor). Without changes in calibre itself (remove No New Privileges flag when spawning the process) there's not way to avoid including Chromium abstraction in the parent profile.
Majority functions I checked worked without issues, any downloads and ebook manipulations I checked functioned as long as ebook library was properly given (calibre by default creates directory with translated name for every language version, but I guessed that for now it's better to leave it to the user as aa-config is already planed, and as far as I understand the concept it seems a proper place to resolve this problem, if at all).
After adding plugin development dirs to local/calibre, even customization functions (generating plugin archive) worked without a problem.
The only thing left is the access to /sys/devices/**/. By default calibre wants everything there to read if it finds a connected device, but i bet that with proper research it could be narrowed down without real issues. Left ToDo comment there.

**Done:**

1. Introducing child profile for poppler apps
2. Fixes for book modification and viewing
3. Full coverage of the binary release
4. Devices access

[Logs_calibre](https://github.com/user-attachments/files/30242224/aa-calibre.txt)

roddhjav:
Thanks, this is good for now. You are right about QtWebEngine.

Merged.

PR #1182, opened 2026-07-23
fix(profiles): allow reading ICC color profiles

last5bits:
Several profiles inherit file descriptors for ICC color profiles from colord/kwin_wayland. Add /usr/share/color/icc/{,**} r, to the X-strict abstraction (covering xwayland, xkbcomp, xprop, xrdb, and KDE profiles that pull in kde-strict/X-strict). Add the rule directly to kwin_wayland's at-spi sub-profile, which does not include X-strict.

Log entries fixed:
```
apparmor="ALLOWED" operation="file_inherit" profile="kcminit" name="/usr/share/color/icc/colord/WideGamutRGB.icc" requested_mask="r" denied_mask="r"
apparmor="ALLOWED" operation="open" profile="kwin_wayland" name="/usr/share/color/icc/colord/WideGamutRGB.icc" requested_mask="r" denied_mask="r"
apparmor="ALLOWED" operation="file_inherit" profile="kwin_wayland//at-spi" name="/usr/share/color/icc/colord/WideGamutRGB.icc" requested_mask="r" denied_mask="r"
```

*AI-assisted*

### DIFF
--- a/apparmor.d/groups/freedesktop/xkbcomp
+++ b/apparmor.d/groups/freedesktop/xkbcomp
@@ -21,6 +21,7 @@ profile xkbcomp @{exec_path} flags=(attach_disconnected) {
   @{exec_path} mr,
 
   /usr/share/X11/xkb/** r,
+  /usr/share/color/icc/{,**} r,
 
   /var/lib/xkb/server-@{int}.xkm w,
   /var/lib/xkb/compiled/server-@{int}.xkm rw,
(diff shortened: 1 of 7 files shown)


roddhjav:
This is not the good abstraction to add this (the abstraction is also used for all desktop app). There is already reference to `/usr/share/color/icc/` in some profiles. So I think this can be added directly in the concerned profiles.

last5bits:
> This is not the good abstraction to add this (the abstraction is also used for all desktop app). There is already reference to `/usr/share/color/icc/` in some profiles. So I think this can be added directly in the concerned profiles.

@roddhjav Fixed.

last5bits:
Added fixes for additional affected profiles, log entries can be found in the commit message.

roddhjav:
Thanks, merged

PR #1185, opened 2026-07-31
feat: add fedora support

CatPieLeaf:
# This PR is CHONKY. >:)

This adds Fedora as a supported distribution, following the same pattern as the existing arch/debian/ubuntu/opensuse support. It also folds in a bunch of contributions from [RakuOS](https://rakuos.org/), an atomic/ostree-based Fedora distro that recently adopted Apparmor and Apparmor.d, since a lot of what they needed overlaps with what any Fedora system needs. This might also help Nobara, since they also use apparmor but without apparmor.d.... basically unprotected.

I've also set-up a COPR with testing builds for anyone to try.
https://copr.fedorainfracloud.org/coprs/catpieleaf/apparmor.d/

It's not perfect and I'll probably keep sending fixes as I find more denials, but the base is solid. Most of what you need for a Fedora desktop to actually work under AppArmor without constant denials is already here. I've been running this on Fedora KDE (not Workstation/GNOME) day to day and iterating on real AVC logs from my own machine, so KDE-specific stuff got a lot more mileage than GNOME did. GNOME should mostly work since a lot of the fixes are generic (dbus, flatpak, xdg-portal stuff), but I haven't hammered on it the way I have KDE.

What's in here is..

Core distro stuff:
- "fedora" recognized as a distribution (it already fell through to the right ID via /etc/os-release, this just wires up the family mapping so #aa:only dnf works)
- Fixed the sudo/sudo-rs double-attachment bug that was previously reverted upstream for Ubuntu - Fedora ships sudo-rs as the real /usr/bin/sudo binary same as recent Ubuntu, and both profiles were claiming that path
- @{sbin} merged-usr tunable extended to cover Fedora (same shape as Arch)
- Fedora KDE spin's default kdeglobals path added
- New dists/ignore/fedora.ignore and dists/flags/fedora.flags following the existing per-distro convention

New profiles:-
- groups/dnf/{dnf,dnf5} - was completely missing before, dnf5 in particular needed a fair amount of work (ca-trust, crypto-policies, sysimage paths, librepo tmp files, gpg subprocess handling)
- A handful of generic app profiles that just didn't exist yet: aria2, doas, heroic, mangoapp, nvme-cli, podman, visudo, yubikey-manager, systemd-run, all came from RakuOS.
- Some KDE/browser profiles that were missing: ark, gwenview, kate, kcalc, kinfocenter, kmenuedit, spectacle, brave-origin, claude-desktop, helium-bin
- Full COSMIC group (36 files) ported over, though it stays behind the existing WIP ignore like upstream intends
- ksshaskpass got its own profile instead of inheriting whatever confined it (was a mess of KDE/Qt/font denials before)
- A specfile for Fedora builds.

A lot of denial fixes across existing profiles..
Way too many to list individually, but the shape of it: KDE apps needed authselect paths, crypto-policies, kde-settings profile dir. Flatpak stuff needed a bunch of sandbox filesystem and dbus work (xdg-desktop-portal, xdg-document-portal, flatpak-portal, flatpak-session-helper). Xorg/X11 tools needed xauth handling cleaned up. mkvtoolnix, smplayer, qbittorrent, okular, nvidia-settings all needed the usual file-picker-touches-your-whole-home-and-every-mount treatment, done through owner rules instead of blanket access. gparted needed cgroup path handling for both plain and containerized sessions. packagekitd needed a lot of dnf5-specific gpg/rpm work. git needed glab support alongside the existing gh handling, and its ssh-askpass path needed fixing.

One thing specific to atomic Fedora: on ostree/bootc systems your real home is at /var/home/<user>, with /home just a symlink - AppArmor matches the resolved path, not through the symlink, so without this every @{HOME}-derived rule in the whole project silently didn't apply. That's a one-line tunable fix but it matters a lot if you're on Silverblue/Kinoite or similar.

Known gaps:
GNOME Workstation hasn't gotten the same real-world testing KDE has, so there's probably denials I haven't hit yet. Same goes for anything outside my own daily driver setup - browsers/apps I don't personally use, hardware I don't have, etc. If something's missing, it's very likely just because I haven't hit that code path myself yet, not because it's intentionally excluded. This is meant to be a starting point to begin supporting Fedora.

Currently we're testing it on Fedora KDE and soon in RakuOS.
Soon we'll let Nobara folks know about it too.

### DIFF
--- a/apparmor.d/abstractions/X-strict
+++ b/apparmor.d/abstractions/X-strict
@@ -32,6 +32,8 @@
   owner @{run}/user/@{uid}/X11/Xauthority r,
   owner @{run}/user/@{uid}/xauth_@{rand6} rl,
 
+  /dev/udmabuf rw,
+
   include if exists <abstractions/X-strict.d>
 
 # vim:syntax=apparmor
--- a/apparmor.d/abstractions/app/bus
+++ b/apparmor.d/abstractions/app/bus
@@ -10,6 +10,10 @@
   include <abstractions/consoles>
   include <abstractions/nameservice-strict>
 
+  # dbus-launch bootstraps X11/dbus auth cookies it doesn't own
+  capability dac_override,
+  capability dac_read_search,
+
   @{bin}/dbus-launch  mrix,
   @{bin}/dbus-send    mrix,
 
--- a/apparmor.d/abstractions/app/flatpak
+++ b/apparmor.d/abstractions/app/flatpak
@@ -96,6 +96,56 @@
   # apply_extra
   /app/extra/** w,
 
+  # Qt runtime translations, remapped inside the flatpak sandbox
+  /usr/translations/{,**} r,
+
+  /etc/issue r,
+  /etc/os-release r,
+
+  # Self-introspection within the sandbox's own low-pid namespace
+  @{PROC}/[0-9]*/ r,
+  @{PROC}/[0-9]*/mountinfo r,
+  @{PROC}/[0-9]*/mounts r,
+  @{PROC}/[0-9]*/net/arp r,
+  @{PROC}/[0-9]*/oom_score r,
+  @{PROC}/[0-9]*/task/[0-9]*/comm rw,
+  @{PROC}/[0-9]*/statm r,
+
+  # Some apps blindly open() with O_CREAT even for files that already
+  # exist, e.g. AnyDesk's /proc/cpuinfo probe
+  @{PROC}/cpuinfo rw,
+
+  # Hardware/platform device identification (e.g. Electron gamepad enumeration)
+  @{sys}/devices/platform/uevent r,
+  @{sys}/devices/platform/**/uevent r,
+
+  # Network interface / PCI device enumeration (e.g. lspci, network client
+  # apps reading a MAC address for machine identification)
+  @{sys}/bus/pci/slots/ r,
+  @{sys}/bus/pci/slots/*/address r,
+  @{sys}/devices/@{pci}/label r,
+  @{sys}/devices/@{pci}/net/*/address r,
+  @{PROC}/bus/pci/devices r,
+
+  # CPU topology/capability enumeration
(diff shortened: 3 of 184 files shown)


roddhjav:
Thanks a lot for this work. It is good to have support for the few fedora based distributions that use AppArmor.

However, and it is kind of obvious, I am not merging a PR that big (with one commit) into the project, I can't ever review it nor test it properly this way. Thus, you will have to split the work into multiple PR. Proper feedback will then be provided in each PR (no offense, but from what I see in the commit, a lot of changes will need to be done).

From a quick review; here are the PR you need to make

1. Start by the core building part. E.g. the changes to the prebuild system.
2. Add test images for the selected Fedora based distribution (https://apparmor.pujol.io/development/vm/). Without tests we simply cannot support a distribution.
3. Add the dnf profiles (and the tests).
4. Some complex profile will need a PR only for them (doas, podman...): I have an early version for a profile for podman (and its deps), however, it has never made to the repo because it is simply not stable enough. In other word, this profile by itself requires a lot of work and test to not break someone else computer. It goes the same way for systemd-run.
5. Same for any changes related to a group of profiles (kde, gnome)
6. Profies changes not really specific to Fedora
7. I am sorry; but as of today I am not adding support for another DE; it is too much work to support. So I won't accept any of the cosmic profiles (a lot of changes seems to be needed to these profiles anyway, but that's fine).

Do all these PRs sequentially, so feedback from a PR can be applied to later PR.

CatPieLeaf:
No problem! I completely understand you!
Also, many thanks for the information and support :)

Soon I'll be splitting this PR into multiple commits and PRs.

I'll update this PR with all the commits that can fit on a single PR. I'll probably start with the prebuild changes, specfile and some go files

Feel free to make any critiques or suggestions! That would be very appreciated and help a lot!

CatPieLeaf:
I'll be splitting this into 13 thematic PRs, because its way too many changes, i can't just make a ton of individual PRs for each thing.

CatPieLeaf:
This main PR will be closed, as main branch won't be used for any PR.

PR #1187, opened 2026-08-02
feat: recognize Fedora as a supported distribution (core prebuild system)

CatPieLeaf:
# Part 1 of 14 -  Bringing Fedora support for apparmor.d
This PR is the first in a series bringing Fedora support upstream, split
into small sequential PRs. Every other PR in the
series depends on this one merging first - none of them build under
`DISTRIBUTION=fedora` without it.

## Summary

Adds Fedora into the prebuild system: dnf as its package family, its own ignore list, its own
tunable overrides, and a fix to the profile-reattachment builder task that
a couple of setuid-PAM-helper profiles (sudo, sudo-rs, su, su-rs,
unix-chkpwd) need regardless of distribution.

## What's in this PR

- `pkg/tasks/os.go`: registers `dnf` → `fedora` in the package-family map.
- `pkg/configure/configure.go`, `pkg/builder/userspace.go`: fedora as a
  recognized `DISTRIBUTION` value; fedora gets the same combined
  `/{,usr/}{,s}bin` sbin definition arch already uses.
- `pkg/builder/attach.go`: excludes sudo/sudo-rs/su/su-rs/unix-chkpwd from
  the disconnected-path reattachment transform - these Px-transition into
  setuid PAM helpers, and reattachment breaks that exec transition under
  real kernel enforcement even though the resulting profile compiles and
  merges cleanly. Not fedora-specific, but discovered while testing fedora.
- `dists/ignore/fedora.ignore` (new) + the other four distros' ignore lists
  (each now also excludes `apparmor.d/groups/dnf`).
- `apparmor.d/tunables/home.d/apparmor.d`: `@{HOMEDIRS}+=/var/home/`,
  gated `#aa:only fedora` - ostree/bootc Fedora Atomic variants (Silverblue,
  Kinoite, IoT, CoreOS) keep real home directories at `/var/home/<user>/`
  with `/home` a symlink; AppArmor matches the resolved path, so `@{HOME}`
  never matches on those variants without this.
- `apparmor.d/tunables/multiarch.d/system`: adds the `*-redhat-linux*`
  multiarch triplet (Fedora/RHEL doesn't use the `-gnu` suffix the other
  distros share).
- `docs/development/directives.md`: documents `fedora`/`dnf` in the
  only/exclude directive reference.
- `dists/apparmor.d-fedora.spec` (new): Fedora RPM spec, sourced from
  GitHub release tags, depending on `apparmor-parser`/`apparmor-utils`/
  `apparmor-profiles` (no `distribution-release`/`golang-packaging` macros
  on Fedora, unlike the existing openSUSE-oriented `dists/apparmor.d.spec`
  this mirrors). Drops the systemd `AppArmorProfile=` drop-ins at install
  time - forcing dbus(-broker).service under confinement that way causes a
  boot-time race that cascades into PAM session bus / kwallet / polkit auth
  breakage on Fedora Atomic variants.

### DIFF
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -3,6 +3,14 @@
 # Copyright (C) 2021-2024 Alexandre Pujol <alexandre@pujol.io>
 # SPDX-License-Identifier: GPL-2.0-only
 
+# ostree/bootc-based Fedora Atomic variants (Silverblue, Kinoite, IoT, CoreOS)
+# keep real user home directories at /var/home/<user>/, with /home a symlink
+# to /var/home - AppArmor matches the resolved path, not through the symlink,
+# so without this, @{HOME} (and everything derived from it: user_config_dirs,
+# user_cache_dirs, etc.) never matches on those variants.
+#aa:only fedora
+@{HOMEDIRS}+=/var/home/
+
 # To allow extended personalisation by the user without breaking everything.
 # All apparmor profiles should always use the variables defined here.
 
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -3,11 +3,8 @@
 # Copyright (C) 2021-2024 Alexandre Pujol <alexandre@pujol.io>
 # SPDX-License-Identifier: GPL-2.0-only
 
-# ostree/bootc-based Fedora Atomic variants (Silverblue, Kinoite, IoT, CoreOS)
-# keep real user home directories at /var/home/<user>/, with /home a symlink
-# to /var/home - AppArmor matches the resolved path, not through the symlink,
-# so without this, @{HOME} (and everything derived from it: user_config_dirs,
-# user_cache_dirs, etc.) never matches on those variants.
+# ostree/bootc Fedora Atomic variants keep homes at /var/home/<user>, /home a
+# symlink; AppArmor matches the resolved path, not through the symlink.
 #aa:only fedora
 @{HOMEDIRS}+=/var/home/
 
--- a/apparmor.d/tunables/home.d/apparmor.d
+++ b/apparmor.d/tunables/home.d/apparmor.d
@@ -3,9 +3,9 @@
 # Copyright (C) 2021-2024 Alexandre Pujol <alexandre@pujol.io>
 # SPDX-License-Identifier: GPL-2.0-only
 
+#aa:only fedora
 # ostree/bootc Fedora Atomic variants keep homes at /var/home/<user>, /home a
 # symlink; AppArmor matches the resolved path, not through the symlink.
-#aa:only fedora
 @{HOMEDIRS}+=/var/home/
 
 # To allow extended personalisation by the user without breaking everything.
--- a/dists/apparmor.d-fedora.spec
+++ b/dists/apparmor.d-fedora.spec
@@ -1,0 +1,86 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (c) 2023 SUSE LLC
+# Copyright (c) 2023 Christian Boltz
+# Copyright (C) 2023-2024 Alexandre Pujol <alexandre@pujol.io>
+# Copyright (c) 2026 CatPieLeaf <catpieleaf@proton.me>
+# SPDX-License-Identifier: GPL-2.0-only
+
+%define _complain 0
+
+Name:           apparmor.d
+Version:        0.4910.0
+Release:        1%{?dist}
+Summary:        Full set of AppArmor policies
+License:        GPL-2.0-only
+URL:            https://github.com/roddhjav/apparmor.d
+Source0:        %{url}/archive/refs/tags/v%{version}.tar.gz
+
+Requires:       apparmor-profiles
+Requires:       apparmor-parser
+Requires:       apparmor-utils
+BuildRequires:  just
+BuildRequires:  golang
+BuildRequires:  systemd-rpm-macros
+
+%description
+AppArmor.d is a set of over 1500 AppArmor profiles whose aim is to
+confine most Linux based applications and processes.
+
+%if %{_complain}
+Built in COMPLAIN mode: violations are logged but not blocked.
+Use this mode with aa-logprof for fixing profiles.
+%else
+Built in ENFORCE mode: violations are actively blocked. Make sure
+you have already validated this profile set in complain mode.
+%endif
+
+%prep
+%autosetup -n %{name}-%{version}
+
+%build
+%if %{_complain} == 1
+just complain
+%else
+just enforce
+%endif
+
+%install
+just destdir="%{buildroot}" install-prebuilt
+just destdir="%{buildroot}" install-base
+just destdir="%{buildroot}" install-tools
+
+# Do not force dbus(-broker).service under AppArmorProfile= via systemd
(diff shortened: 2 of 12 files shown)


apparmor.d/tunables/home.d/apparmor.d
```diff
@@ -3,6 +3,14 @@
 # Copyright (C) 2021-2024 Alexandre Pujol <alexandre@pujol.io>
 # SPDX-License-Identifier: GPL-2.0-only
 
+# ostree/bootc-based Fedora Atomic variants (Silverblue, Kinoite, IoT, CoreOS)
```
roddhjav: Move the aa:only directive before the comment such as on non Fedora system; we don't end up with a comment without related variable;

Also, you need to teach your LLM to be more concise when writhing comments. I have no issue with LLM for such a tasks, however, this comments is simply too long/complex for what it is. For instance, here, the following is enough:

> ostree/bootc based Fedora Atomic variants keep real user home directories at /var/home/<user>/, with /home a symlink to /var/home.

dists/apparmor.d-fedora.spec
```diff
@@ -0,0 +1,86 @@
...
+just destdir="%{buildroot}" install-prebuilt
+just destdir="%{buildroot}" install-base
+just destdir="%{buildroot}" install-tools
+
+# Do not force dbus(-broker).service under AppArmorProfile= via systemd
+# drop-in. dbus underlies PAM's session bus, kwallet registration, and
+# polkit, so a boot-time dbus confinement race cascades into hard-to-diagnose
+# auth/wallet breakage on Fedora Atomic variants.
+rm -rf %{buildroot}/usr/lib/systemd/system/*.service.d
+rm -rf %{buildroot}/usr/lib/systemd/user/*.service.d
```
roddhjav: What is the problem with AppArmorProfile= ?

The dbus profiles have been designed to work this when. Therefore, not using this would break all the dbus profiles implementation (thus, would break the whole system)

pkg/builder/attach.go
```diff
@@ -26,6 +26,20 @@ func NewAttach() *ReAttach {
 	}
 }
 
+// reAttachExclude lists profiles that must keep their plain attach_disconnected
+// flag without the attach_disconnected.path=@{att} extension or the attached/*
+// abstraction swap. These profiles Px-transition into small setuid PAM helpers
+// (unix_chkpwd) to authenticate the user, and the disconnected-path reattachment
+// mechanism breaks that exec transition under real kernel enforcement even though
+// it compiles and merges cleanly.
```
roddhjav: I don't really understand the problem here. We never had such issue on other distribution.

roddhjav:
As Fedora does not support Apparmor, can you explain how you get Apparmor working? This will be useful later, as a test image with Fedora (configured for apparmor) will have to be provided (see https://github.com/roddhjav/apparmor.d/tree/main/tests/cloud-init and https://github.com/roddhjav/apparmor.d/tree/main/tests/packer)

I suppose you use a different kernel. Are you aware you also need to build systemd and dbus-broker with support for apparmor? Other software such as docker, libvirt need the same kind of work too.

CatPieLeaf:
Yup! To make it work on fedora..
I'm currently using my own custom kernel (p03-kernel) with apparmor support enabled.
Then i just disabled selinux on grub cmdline and used apparmor packages from terrapkg.
And yes, I'm using a modified systemd (from rakuos repo).
But some time ago i also managed to load apparmor without a modified systemd, by just creating a service to start it.
im using stock dbus-broker from fedora.

Soon I'll check your feedback and answer everything 

it works perfectly so far, except for a few denials that can be easily fixed.

dists/apparmor.d-fedora.spec
CatPieLeaf: This came from RakuOS's apparmor.d specfile, when they tried to ship their own fork of apparmor.d. the full explanation in my other specfile says..

```
# Do not force dbus(-broker).service under AppArmorProfile= via systemd
# drop-in. RakuOS (a Fedora-based distro shipping this same profile set)
# tried this repeatedly and gave up on it: "still causing boot/service
# problems after repeated re-enable attempts" (their git history has several
# enable/disable round-trips ending in permanently disabled). dbus underlies
# PAM's session bus, kwallet registration, and polkit, so a boot-time dbus
# confinement race cascades into exactly this kind of hard-to-diagnose
# auth/wallet breakage.
```

pkg/builder/attach.go
CatPieLeaf: During my tests, sudo-rs and sudo profiles were going through some processing during build, and this was causing sudo to always fail or complain about incorrect password. Also checkpwd was failing too.

Sudo failures have never occured on rakuos apparmor.d fork because they manually copied the profiles as-is instead of using go and justfile to process the profiles.
i can show you their old specfile if you'd like.

apparmor.d/tunables/home.d/apparmor.d
CatPieLeaf: Alright! soon i'll address this.

CatPieLeaf: Done!

dists/apparmor.d-fedora.spec
roddhjav: This does not explain the root cause of the issue. Furthermore, it is a core part of the project, thus it simply cannot be disabled. Remove this for now, we will investigate this later (using the test image).

pkg/builder/attach.go
roddhjav: This should not be related to attach_disconnected.path (or simply some rules are missing and should be added). Remove this for now, we will investigate this later (using the test image).

roddhjav:
> I'm currently using my own custom kernel (p03-kernel) with apparmor support enabled.

As you will need (for the next PR) to provide a test image for fedora, this kernel will have to be present in it (or RakuOS's kernel as the test VM image need to be supported). Same for systemd and dbus-broker.

> But some time ago i also managed to load apparmor without a modified systemd, by just creating a service to start it. im using stock dbus-broker from fedora.

For apparmor this would work fine yes. However; in apparmor.d you must use early policy load. A feature from systemd that load the compiled policies into the kernel at a very early stage (way before the first service is started). It is the only solution to ensure all profiles from apparmor.d are loaded early enough. And for this, you need systemd compiler with the good flag.

dists/apparmor.d-fedora.spec
CatPieLeaf: done :)

pkg/builder/attach.go
CatPieLeaf: removed

CatPieLeaf:
Done! :)
i addressed all the reviews, now i'm gonna open a PR with VM images.

they'll have RakuOS repo for apparmor and systemd packages, and p03 kernel.
also, RakuOS uses p03 kernel as their main kernel, there's no rakuos kernel.

after removing `rm -rf %{buildroot}/usr/lib/systemd/system/*.service.d`
`rm -rf %{buildroot}/usr/lib/systemd/user/*.service.d` from specfile, i went into a black screen before window manager while using apparmor packages from terrapkg. This was fixed by swapping to apparmor packages from rakuos.

dbus-broker service doesn't work properly with terrapkg apparmor packages, which explains the issues with those, i'll look into that with terra folks later (their packages have missing apparmor.service)

roddhjav:
Thanks, merged. We will fix this issues in future PR.

PR #1191, opened 2026-08-07
test(vm): add Fedora 44 VM test images

CatPieLeaf:
# Part 2 of 14 — Fedora support for apparmor.d

Adds Fedora 44 server/GNOME/KDE cloud-init configs and a packer distro-map
entry, modeled file-for-file on the existing archlinux/debian/opensuse
patterns in `tests/cloud-init/` and `tests/packer/`. Also carries a couple
of small fixes to the shared `tests/packer/` provisioning/cleanup scripts
that Fedora's boot flow exposed but that apply to any distro image with
similar needs.

## What's in this PR

- `tests/cloud-init/fedora.yml`: shared package anchors for
  server/gnome/kde flavors.
- `tests/cloud-init/fedora44-{server,gnome,kde}.user-data.yml`.
- `tests/packer/variables.pkr.hcl`: Fedora 44 Cloud Base Generic qcow2 +
  its PGP-signed CHECKSUM file (verified reachable at
  download.fedoraproject.org).
- `tests/packer/builds.pkr.hcl`: split the "wait for cloud-init" step into
  its own provisioner with `expect_disconnect = true`, and added a 30s
  `pause_before` on the provisioner that follows.
- `tests/packer/clean.sh`: guard the `/var/lib/dbus/machine-id` truncate
  behind an existence check.

Four things these images need that don't apply to any other distro's
image in this repo:

1. Fedora ships no apparmor package at all (SELinux is the default LSM).
   The AppArmor userspace stack (apparmor-libs, apparmor-parser,
   apparmor-utils, apparmor-profiles, python3-LibAppArmor, python3-apparmor)
   plus an AppArmor-patched systemd (systemd, systemd-udev, systemd-libs)
   are installed from Rakuos (repo.rakuos.org/rakuos.repo), scoped to
   `--repo rakuos install <names>` - never `distro-sync` or a blanket
   upgrade against that repo, only the named packages.
2. The stock Fedora kernel has no AppArmor LSM built in. These images
   install kernel-p03 (RakuOS's default kernel) and reboot into
   it (`power_state: {mode: reboot, delay: 1}` at the end of each
   user-data file) before the grubby kernel-args step means anything. The
   `delay: 1` matters: cloud-init's default `delay: now` reboots the VM
   right as Packer's wait-loop provisioner detects `boot-finished`, racing
   Packer's own cleanup of that provisioner's temp script and failing the
   build with "Timeout during SSH handshake" - a 1-minute delay gives that
   cleanup (and the provisioner split above) room to finish first.
3. Fedora Cloud images use BLS (BootLoaderSpec), not a monolithic
   `grub.cfg` regenerated from `/etc/default/grub` - `grubby
   --update-kernel=ALL` is used instead of `grub2-mkconfig`.
4. Fedora's dbus-broker doesn't maintain `/var/lib/dbus/machine-id` the
   way Debian/Arch/openSUSE do, so `clean.sh`'s unconditional truncate of
   that path fails there - now guarded with an existence check.

### DIFF
--- a/dists/docker.sh
+++ b/dists/docker.sh
@@ -117,6 +117,25 @@ build_in_docker_rpm() {
 	mv "$VOLUME/$PKGNAME/$OUTPUT/$PKGNAME"*.rpm "$OUTPUT"
 }
 
+build_in_docker_fedora() {
+	local img="$PREFIX""fedora"
+
+	if _exist "$img"; then
+		if ! _is_running "$img"; then
+			_start "$img"
+		fi
+	else
+		docker pull registry.fedoraproject.org/fedora:latest
+		docker run -tid --name "$img" --volume "$VOLUME:$BUILDIR" \
+			registry.fedoraproject.org/fedora:latest
+		docker exec "$img" dnf5 install -y rpm-build golang just systemd-rpm-macros curl jq
+	fi
+
+	docker exec --workdir="$BUILDIR/$PKGNAME" "$img" \
+		sh -c "cp dists/$PKGNAME-fedora.spec dists/$PKGNAME.spec && just build-rpm"
+	mv "$VOLUME/$PKGNAME/$OUTPUT/$PKGNAME"*.rpm "$OUTPUT"
+}
+
 main() {
 	case "$DISTRIBUTION" in
 	archlinux)
@@ -132,6 +151,11 @@ main() {
 	opensuse)
 		sync
 		build_in_docker_rpm "$DISTRIBUTION"
+		;;
+
+	fedora)
+		sync
+		build_in_docker_fedora
 		;;
 
 	*) ;;
--- a/tests/cloud-init/fedora.yml
+++ b/tests/cloud-init/fedora.yml
@@ -83,10 +83,3 @@ kde-packages: &kde-packages
 #  - reboots into kernel-p03 at the end (power_state:, see each
 #    *-user-data.yml) so later apparmor_parser/aa-log checks run against a
 #    kernel that actually has the LSM.
-
-# Set some bash aliases
-setup-bash-aliases: &setup-bash-aliases
-  path: /etc/skel/.bashrc
-  append: true
-  content: |
-    [[ -f ~/.bash_aliases ]] && source ~/.bash_aliases
--- a/tests/cloud-init/fedora.yml
+++ b/tests/cloud-init/fedora.yml
@@ -69,12 +69,12 @@ kde-packages: &kde-packages
 # packages: config module runs before runcmd's final stage, and a YAML
 # sequence-alias can't be spliced into another sequence without becoming an
 # invalid nested list - so each *-user-data.yml below repeats it. It:
-#  - swaps in Rakuos's systemd (systemd, systemd-udev, systemd-libs) - it's
-#    built with AppArmor support, Fedora's stock systemd is not;
 #  - installs the AppArmor userspace stack (apparmor-libs, apparmor-parser,
 #    apparmor-utils, apparmor-profiles, python3-LibAppArmor, python3-apparmor)
-#    from Rakuos (repo.rakuos.org) - Fedora ships no official apparmor
+#    from Terra (repos.fyralabs.com) - Fedora ships no official apparmor
 #    package at all (SELinux is the default LSM);
+#  - swaps in Rakuos's systemd (systemd, systemd-udev, systemd-libs) - it's
+#    built with AppArmor support, Fedora's stock systemd is not;
 #  - installs kernel-p03 from catpieleaf/kernel-p03 (COPR) - the stock
 #    Fedora kernel has no AppArmor LSM built in at all, this custom kernel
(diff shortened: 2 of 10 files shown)


tests/packer/builds.pkr.hcl
```diff
@@ -66,13 +66,24 @@ build {
     ]
   }
 
-  # Full system provisioning
+  # Wait for cloud-init to finish. Some distros (Fedora, to switch onto
+  # kernel-p03) reboot as part of their user-data's own power_state: right
+  # as boot-finished appears, killing this SSH session - expect_disconnect
+  # lets that through instead of failing the build; distros that don't
+  # reboot here just finish this step normally.
```
roddhjav: Again, you are not writhing a book. *Wait for cloud-init to finish.* is enough.

tests/cloud-init/fedora44-kde.user-data.yml
```diff
@@ -0,0 +1,44 @@
...
+  - dnf5 -y copr enable catpieleaf/kernel-p03
+  - dnf5 -y install kernel-p03
+
+  # Build and install apparmor.d itself from this fork's main branch,
+  # locally - not from the copr, so the VM tests current local work
+  - git clone --depth 1 -b main https://github.com/CatPieLeaf/apparmor.d-fedora.git /root/apparmor.d-fedora
+  - |
+    cd /root/apparmor.d-fedora
+    cp dists/apparmor.d-fedora.spec dists/apparmor.d.spec
+    DISTRIBUTION=fedora just rpm
```
roddhjav: This is conceptually wrong. We do not want to build and test your own apparmor.d-fedora repo (btw, that could not be done as root user anyway). We want to build and install this very repo to test and develop it.

The build part is done through `just build-rpm` , itself build though a container as a first step of `just img fedora 44 kde`.

Furthermore, the install part of this package must be done outside of cloud init. In the `init.sh` script.

tests/cloud-init/fedora.yml
```diff
@@ -0,0 +1,95 @@
...
+#  - reboots into kernel-p03 at the end (power_state:, see each
+#    *-user-data.yml) so later apparmor_parser/aa-log checks run against a
+#    kernel that actually has the LSM.
+
+# Set some bash aliases
+setup-bash-aliases: &setup-bash-aliases
+  path: /etc/skel/.bashrc
+  append: true
+  content: |
+    [[ -f ~/.bash_aliases ]] && source ~/.bash_aliases
```
roddhjav: Move this to common.yml, and remove the one from archlinux.yml

tests/cloud-init/fedora44-gnome.user-data.yml
```diff
@@ -0,0 +1,44 @@
+#cloud-config
+
+packages: *gnome-packages
+
+runcmd:
+  # AppArmor userspace tools from Rakuos (no official Fedora package)
+  - dnf5 -y config-manager addrepo --from-repofile=https://repo.rakuos.org/rakuos.repo
+  # Rakuos systemd is built with AppArmor support, Fedora's stock one is not
+  - dnf5 -y --repo rakuos install systemd systemd-udev systemd-libs
```
roddhjav: To force the use of rakuos repo, here you need:
```suggestion
  - dnf5 -y --repo rakuos distro-sync "systemd*"
```

Otherwise, the generated VM still uses Fedora's systemd.

tests/cloud-init/fedora44-kde.user-data.yml
CatPieLeaf: Oops, that wasn't supposed to go into the PR. I only added my fork temporarily for testing vm images locally before making the PR.
Thanks for pointing out, I'll remove it later

tests/packer/builds.pkr.hcl
CatPieLeaf: Done :+1:

tests/cloud-init/fedora44-kde.user-data.yml
CatPieLeaf: Done

dists/docker.sh
```diff
@@ -117,6 +117,25 @@ build_in_docker_rpm() {
 	mv "$VOLUME/$PKGNAME/$OUTPUT/$PKGNAME"*.rpm "$OUTPUT"
 }
 
+build_in_docker_fedora() {
```
roddhjav, line 120: FYI, in theory this should be the same function (and build system in general) than for opensuse, as it targets rpm. For now, I am okay having a different function (and spec file) to easily hack on it while we add support for fedora.

The rest of the function should use $BASEIMAGE as base image.

tests/cloud-init/fedora44-gnome.user-data.yml
roddhjav: This is still not changed (and it needs to be applied on all images).

roddhjav:
Thanks, I merged it with minor modifications.

PR #1192, opened 2026-08-07
add profile for qutebrowser

valoq:
Another browser with chromium engine in the backend via QtWebEngine

### DIFF
--- a/apparmor.d/profiles-m-r/qutebrowser
+++ b/apparmor.d/profiles-m-r/qutebrowser
@@ -1,0 +1,55 @@
+# apparmor.d - Full set of apparmor profiles
+# Copyright (C) 2026 Alexandre Pujol <alexandre@pujol.io>
+# SPDX-License-Identifier: GPL-2.0-only
+
+abi <abi/5.0>,
+
+include <tunables/global>
+
+@{domain} = org.chromium.Chromium
+
+@{exec_path} = @{bin}/qutebrowser
+profile qutebrowser @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/common/chromium>
+  include <abstractions/desktop>
+  include <abstractions/graphics>
+  include <abstractions/nameservice-strict>
+  include <abstractions/python>
+  include <abstractions/ssl_certs>
+  include <abstractions/user-download-strict>
+
+  network inet dgram,
+  network inet stream,
+  network inet6 dgram,
+  network inet6 stream,
+  network netlink raw,
+
+  signal (send receive) set=term peer=qutebrowser,
+
+  ptrace (read readby) peer=qutebrowser,
+
+  unix (bind listen) type=stream addr=@{run}/user/@{uid}/qutebrowser/ipc-*,
+
+  @{exec_path} mrix,
+
+  @{python_path} rix,
+  @{lib}/qt6/{,libexec/}QtWebEngineProcess ix,
+
+  owner @{user_config_dirs}/chromium/WidevineCdm/** mr,
+  owner @{user_config_dirs}/qutebrowser/ rw,
+  owner @{user_config_dirs}/qutebrowser/** rwlk -> @{user_config_dirs}/qutebrowser/**,
+
+  owner @{user_cache_dirs}/qutebrowser/ w,
+  owner @{user_cache_dirs}/qutebrowser/** rw,
+
+  owner @{user_share_dirs}/qutebrowser/ rw,
+  owner @{user_share_dirs}/qutebrowser/** rwlk -> @{user_share_dirs}/qutebrowser/**,
+
+  owner @{run}/user/@{uid}/qutebrowser/ w,
+  owner @{run}/user/@{uid}/qutebrowser/ipc-* w,
+
+  include if exists <local/qutebrowser>
+}
+
+# vim:syntax=apparmor


apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,65 @@
...
+
+include <tunables/global>
+
+@{exec_path} = @{bin}/qutebrowser
+profile qutebrowser @{exec_path} {
+  include <abstractions/base>
+  include <abstractions/nameservice-strict>
+  include <abstractions/fonts>
+  include <abstractions/python>
+  include <abstractions/qt6>
```
roddhjav: `fonts` and `qt6` are already included in `desktop`

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,65 @@
...
+  include <abstractions/desktop>
+  include <abstractions/ssl_certs>
+  include <abstractions/vulkan>
+  include <abstractions/opengl>
```
roddhjav: Instead of those, use `graphics`

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,65 @@
...
+  network inet6 stream,
+  network inet6 dgram,
+  network netlink raw,
+
+  signal (send receive) set=term peer=qutebrowser,
+  ptrace (read readby) peer=qutebrowser,
+
+  unix (bind listen) type=stream addr=@{run}/user/@{uid}/qutebrowser/ipc-*,
+
+  @{exec_path} ixr,
```
roddhjav: This should be `  @{exec_path} mrix,`

roddhjav:
Because qutebrowser is based on chromium, you should either:
- Directly use app/chromium if it is a fork of chromium
- Use common/chromium, if it is not a fork but uses the same tech.

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,65 @@
...
+  owner link @{user_config_dirs}/qutebrowser/quickmarks -> @{user_config_dirs}/qutebrowser/#{int8},
+
+  owner @{user_cache_dirs}/qutebrowser/ w,
+  owner @{user_cache_dirs}/qutebrowser/** rw,
+
+  owner @{user_share_dirs}/qutebrowser/ w,
+  owner @{user_share_dirs}/qutebrowser/** rwk,
+  owner link @{user_share_dirs}/qutebrowser/sessions/_autosave.yml -> @{user_share_dirs}/qutebrowser/sessions/#*,
+
+  owner @{HOME}/.pki/nssdb/** r,
```
roddhjav: Use `abstractions/nss`. However, it will be handled by common/chromium

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,65 @@
...
+
+  owner /tmp/.org.chromium.Chromium.* rw,
+  owner /dev/shm/.org.chromium.Chromium.* rw,
```
roddhjav: Will be handled by common/chromium

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -9,14 +9,11 @@ include <tunables/global>
 @{exec_path} = @{bin}/qutebrowser
 profile qutebrowser @{exec_path} {
   include <abstractions/base>
-  include <abstractions/nameservice-strict>
-  include <abstractions/fonts>
-  include <abstractions/python>
-  include <abstractions/qt6>
+  include <abstractions/common/chromium
```
roddhjav: This is not enough, a lot of rules in this profile are already covered by this abstraction, you need to remove them.

Btw, you have a missing `>` at the end of the abs.

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,62 @@
...
+
+  unix (bind listen) type=stream addr=@{run}/user/@{uid}/qutebrowser/ipc-*,
+
+  @{exec_path} mrix,
+  @{lib}/qt6/QtWebEngineProcess ix,
+
+  owner @{user_config_dirs}/qutebrowser/ w,
+  owner @{user_config_dirs}/qutebrowser/** rw,
+  owner @{user_config_dirs}/chromium/WidevineCdm/** mr,
+  owner @{user_config_dirs}/user-dirs.dirs r,
```
roddhjav: This is already in the desktop abs.

apparmor.d/profiles-m-r/qutebrowser
```diff
@@ -0,0 +1,62 @@
...
+  owner link @{user_config_dirs}/qutebrowser/bookmarks/urls -> @{user_config_dirs}/qutebrowser/bookmarks/#{int8},
+  owner link @{user_config_dirs}/qutebrowser/quickmarks -> @{user_config_dirs}/qutebrowser/#{int8},
```
roddhjav: Simplify this a bit. It is qutebrowser own directory, it can do anything in it:
```
  owner @{user_config_dirs}/qutebrowser/ rw,
  owner @{user_config_dirs}/qutebrowser/** rwlk -> @{user_config_dirs}/qutebrowser/**,
```

This will save use some maintenance for when they will rename files; and it has no security consequence.

This also applies to user_share_dirs.

apparmor.d/profiles-m-r/qutebrowser
roddhjav: So, now you need to remove this rule as it is already handled by common/chromium

apparmor.d/profiles-m-r/qutebrowser
roddhjav: So, now you need to remove this rule as it is already handled by common/chromium

PR #1193, opened 2026-08-08
re-add Terrapkg sample-profile ignore entries

CatPieLeaf:
dig/nslookup/podman/free conflict with Terrapkg's apparmor-parser package's own same-named bundled sample profiles (different content, RPM file conflict). Restored from main, where this same revert was previously dropped.

### DIFF
--- a/dists/ignore/fedora.ignore
+++ b/dists/ignore/fedora.ignore
@@ -1,3 +1,10 @@
+# Same-named sample profiles bundled by Terrapkg's apparmor-parser package
+# under /etc/apparmor.d/ (different content), causing RPM file conflicts
+apparmor.d/profiles-a-f/dig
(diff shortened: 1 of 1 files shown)


roddhjav:
These are already ignored as it uses apparmor 5: 
https://github.com/roddhjav/apparmor.d/blob/74be74665765f82508ad0735dcbfbb307e4d6869/pkg/configure/configure.go#L119-L135

CatPieLeaf:
Nah, without this it shows file conflict error on dnf install.

CatPieLeaf:
Screenshot_20260808_034749

look what happens when this is removed.

CatPieLeaf:
@roddhjav

roddhjav:
That's an issue in your build, with apparmor 5, the files should not be presented in the apparmor.d package

PR #1204, opened 2026-08-26
Add terminfo abstraction to mtr

hello-smile6:
mtr is capable of working both as a graphical and terminal-based application, and only is built as a graphical one when using the mtr (rather than mtr-tiny) package on Debian or in other similar scenarios, being explicitly not built as a graphical application on Arch. Therefore, the terminfo abstraction is necessary for mtr to work in non-graphical (by various meanings of the word) environments.


roddhjav:
Thanks, merged!

PR #1205, opened 2026-08-29
feat(firefox) introducing firefox-gfxtest profile

CossackLucas:
Firefox is merging all of the graphics related test on Linux into one binary with version 155 (stable release should be on Sep 1st):
https://bugzilla.mozilla.org/show_bug.cgi?id=1949093
I've used this profile for a while with Nightly and tested it later on VMs (KDE and Gnome, Arch and OpenSuse) and it seems fine, but for a few weeks I had problems with Ubuntu VMs, so those are left open. For this reason I gave it forced complain flag.
The rest of the test binaries were removed from packages, so there could be a chance to remove 3 separated profiles after ESR switches to being based on version 155+ and/or distros ditch the previous versions, but for now I haven't touched them.

Logs from tests:
[aa-firefox-gfxtest-merged.txt](https://github.com/user-attachments/files/31588086/aa-firefox-gfxtest-merged.txt)


roddhjav:
Thanks, merged

PR #1212, opened 2026-08-31
Improve execution of nmap

hello-smile6:
This PR allows nmap to use additional types of network sockets and access data it wants, and should specifically help with spoofing source IPs during a privileged UDP scan over an IP-based VPN or physical network interface (e.g. tun or wireguard).


roddhjav:
Thanks

PR #1213, opened 2026-08-31
Fix issues with networkd

hello-smile6:
Allow networkd to do qdisc configuration and access DHCP lease data.


roddhjav:
THanks, merged

PR #1215, opened 2026-08-31
Allow console for whois

hello-smile6:
On Arch Linux, whois silently fails (doesn't write anything to the console) without the consoles abstraction.


roddhjav:
Thanks, merged.

PR #1219, opened 2026-09-01
Allow tailscale to access consoles

hello-smile6:
Tailscale's CLI fails to write anything to stdout or stderr when prevented from accessing consoles.


roddhjav:
Thanks, merged!

PR #1235, opened 2026-09-20
fix(vm) fix for ArchLinux broken VM builds

CossackLucas:
Removal of these packages was breaking completely systemd (all are hard dependencies) and newly built VMs weren't working, max achievable was getting into sddm without being able to log into or even switching to the TTY. What was the reason for removing those?


roddhjav:
Good catch, thanks, merged!

This is a local only change that should have not been commited: I use a dev version of dbus to have support for dbus mediation.

